PR #1, opened 2020-10-17
Fix formatting

nickygerritsen:
I wanted to add my webhook update, but noticed the formatting is.... less than ideal for code blocks and URL's, so I fixed those.

I hope I got all of it, it was quite a manual process.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -59,47 +59,47 @@ indicate the relative paths of API endpoints with respect to a
 We follow standard REST practices so that a whole collection can be
 requested, e.g. at the URL path
 
-` GET `<https://example.com/api/contests/wf14/teams>
+` GET `<https://example.com/api/contests/wf14/teams>
 
 while an element with specific ID is requested as
 
-` GET `<https://example.com/api/contests/wf14/teams/10>
+` GET `<https://example.com/api/contests/wf14/teams/10>
 
 A collection is always returned as a JSON list of objects. Every object
 in the list represents a single element (and always includes the ID).
 When requesting a single element the exact same object is returned. E.g.
 the URL path
 
-`GET baseurl/`<collection>
+`GET baseurl/`<collection>
 
 returns
 
-`[ { "id":`<id1>`, `<element specific data for id1>`},`  
-`  { "id":`<id2>`, `<element specific data for id2>`},`  
-`     ...`  
+`[ { "id":`<id1>`, `<element specific data for id1>`},`  
+`  { "id":`<id2>`, `<element specific data for id2>`},`  
+`     ...`  
 `]`
 
 while the URL path
 
-`GET baseurl/`<collection>`/`<id1>
+`GET baseurl/`<collection>`/`<id1>
 
 returns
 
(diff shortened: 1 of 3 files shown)


Contest_Archive_Format.md
```diff
@@ -138,17 +140,19 @@ data.
 
 ### Design principles
 
-Endpoints are stored in a single <endpoint>.json file containing the
+Endpoints are stored in a single \<endpoint>.json file containing the
```
RagnarGrootKoerkamp, line 143:
```suggestion
Endpoints are stored in a single `<endpoint>.json` file containing the
```

Contest_Archive_Format.md
```diff
@@ -138,17 +140,19 @@ data.
...
 full list of objects. This file is identical to the API call to
-/<endpoint>. If there are file references in the JSON file these are
+/\<endpoint>. If there are file references in the JSON file these are
```
RagnarGrootKoerkamp, line 145:
```suggestion
`/<endpoint>`. If there are file references in the JSON file these are
```

Wouldn't is be nicer (apart from the effort) to make all endpoints be `code`, so that the `<>` render without issues?

Contest_API.md
```diff
@@ -257,24 +259,26 @@ interpreted relative to the **baseurl** of the API. For example, if
 **baseurl** is <https://example.com/api>, then the following are
 equivalent JSON response snippets pointing to the same location:
 
-` "href":"`<https://example.com/api/contests/wf14/submissions/187/files>`"`  
-` "href":"contests/wf14/submissions/187/files"`
+` "href":"https://example.com/api/contests/wf14/submissions/187/files"`
```
RagnarGrootKoerkamp, line 262: Why are there leading spaces here?

Contest_Archive_Format.md
nickygerritsen: If we are doing this, we need a lot of place to do this. So should we? Then I will update them all.

Contest_Archive_Format.md
nickygerritsen: Might be a good idea. Should I do this?

RagnarGrootKoerkamp (approved):
lgtm

deboer-tim (approved):

eldering (approved):

Contest_Control_System_Requirements.md
```diff
@@ -1300,33 +1300,35 @@ A sequence of mappings with the following keys:
...
+    
+  - letter:     C  
+    short-name: biobots  
+    color:      green  
+    rgb:        '#00ff00'  
+    
+  - letter:     D  
+    short-name: castles  
+    color:      blue  
+    rgb:        '#0000ff'
```
pehrsoderman, line 1325: This is not blue enough for me.

pehrsoderman (changes requested):
Blue should be blue.

Contest_Control_System_Requirements.md
deboer-tim: #0000FF maybe?

PR #4, opened 2020-10-19
Removed references to ACM from CCS Spec

deboer-tim:


Contest_Control_System_Requirements.md
```diff
@@ -7,7 +7,7 @@ permalink: /ccs_system_requirements
 
 This document specifies requirements for operation, verification and
 validation of Programming Contest Control Systems that wish to be
-considered for managing the operation of the [ACM International
+considered for managing the operation of the [International
 Collegiate Programming Contest](http://icpc.baylor.edu) [World
```
niemela, line 11: Link should now be https://icpc.global

niemela (approved):

clevengr (approved):

deboer-tim:
If one of the three who has access to merge is the second approver, may as well just merge it at the same time too?

eldering:
> If one of the three who has access to merge is the second approver, may as well just merge it at the same time too?

Indeed. But I was hesitant to merge this since @niemela still had an unanswered comment. That can still be addressed later though...

Contest_Control_System_Requirements.md
deboer-tim: Sorry, missed this. I will fix via a future PR.

PR #7, opened 2020-10-26
Update webhooks part with proposed changes.

nickygerritsen:
As promised, my proposed changes to the webhooks part.

Probably we want to discuss this next time.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1571,65 +1571,74 @@ If the client fails to respond to multiple requests over a period of
 time (configured for each contest), it will be assumed deactivated and
 automatically removed from future callbacks.
 
-The following endpoint is associated with the webhook:
+The following endpoints is associated with the webhook:
 
-| Endpoint   | Mime-type        | Required? | Source @WF | Description                            |
-| ---------- | ---------------- | --------- | ---------- | -------------------------------------- |
-| `/webhook` | application/json | yes       | CCS        | List or register for webhook callbacks |
+| Endpoint         | Mime-type        | Required? | Source @WF | Description                                                                                                          |
+| ---------------- | ---------------- | --------- | ---------- | -------------------------------------------------------------------------------------------------------------------- |
+| `/webhooks`      | application/json | yes       | CCS        | JSON array of all webhook callbacks with elements as defined in the table below. Also used to register new webhooks. |
+| `/webhooks/<id>` | application/json | yes       | CCS        | JSON object of a single webhook callback with elements as defined in the table below                                 |
 
-JSON elements of webhook objects:
+JSON elements of webhook callback objects:
 
-| Name | Type   | Required? | Nullable? | Description           |
-| ---- | ------ | --------- | --------- | --------------------- |
-| url  | string | yes       | no        | The url for callbacks |
+| Name       | Type            | Required? | Nullable? | Description                                                                           |
+| ---------- | --------------- | --------- | --------- | ------------------------------------------------------------------------------------- |
+id           | ID              | yes       | no        | identifier of the webhook.                                                            |
+url          | string          | yes       | no        | The URL to post HTTP callbacks to.                                                    |
+endpoints    | array of string | yes       | no        | Names of endpoints to receive callbacks for. Empty array means all endpoints.         |
+contest\_ids | array of ID     | yes       | no        | ID’s of contests to receive callbacks for. Empty array means all configured contests. |
+active       | boolean         | yes       | no        | Whether the webhook is enabled.                                                       |
 
-TODO: include filter details?
+The idea behind the `active` field is that if the CCS decides a webhook isn’t behaving and wants to disable sending callbacks to it,
+it can update this active field. It also allows one to temporarily disable a webhook using some UI if a CCS implements it.
 
 ##### Adding a webhook
 
-To register a webhook, you need to post your server's callback url. The
-general format to register a webhook is:
+To register a webhook, you need to post your server's callback URL.
+To do so, perform a `POST` request with a JSON body with the fields (except `id`) from the above table to the `/webhooks` endpoint together with one additional field,
+called `token`. In this field put a client-generated token that can be used to verify that callbacks come from the CCS.
 
-```json
-{"url": "<callback url>", "auth": ... }
+```note
+TODO: What about using sane defaults for `endpoints`, `contest_ids` and `active` when adding a new webhook?
 ```
-
-| Name | Type   | Required? | Nullable? | Description           |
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1571,65 +1571,74 @@ If the client fails to respond to multiple requests over a period of
...
+JSON elements of webhook callback objects:
 
-| Name | Type   | Required? | Nullable? | Description           |
-| ---- | ------ | --------- | --------- | --------------------- |
-| url  | string | yes       | no        | The url for callbacks |
+| Name       | Type            | Required? | Nullable? | Description                                                                           |
+| ---------- | --------------- | --------- | --------- | ------------------------------------------------------------------------------------- |
+id           | ID              | yes       | no        | identifier of the webhook.                                                            |
+url          | string          | yes       | no        | The URL to post HTTP callbacks to.                                                    |
+endpoints    | array of string | yes       | no        | Names of endpoints to receive callbacks for. Empty array means all endpoints.         |
```
niemela, line 1587: How are endpoints specified? Which endpoints are allowed? I.e. what are the possible values here?

Sam as the old list for the event feed? (When did we remove that one? I don't remember)

Contest_API.md
```diff
@@ -1571,65 +1571,74 @@ If the client fails to respond to multiple requests over a period of
...
+contest\_ids | array of ID     | yes       | no        | ID’s of contests to receive callbacks for. Empty array means all configured contests. |
+active       | boolean         | yes       | no        | Whether the webhook is enabled.                                                       |
 
-TODO: include filter details?
+The idea behind the `active` field is that if the CCS decides a webhook isn’t behaving and wants to disable sending callbacks to it,
+it can update this active field. It also allows one to temporarily disable a webhook using some UI if a CCS implements it.
```
niemela, line 1592: This would imply that you might want to have a webhook on the webhook endpoint?

Contest_API.md
```diff
@@ -1571,65 +1571,74 @@ If the client fails to respond to multiple requests over a period of
...
-To register a webhook, you need to post your server's callback url. The
-general format to register a webhook is:
+To register a webhook, you need to post your server's callback URL.
+To do so, perform a `POST` request with a JSON body with the fields (except `id`) from the above table to the `/webhooks` endpoint together with one additional field,
+called `token`. In this field put a client-generated token that can be used to verify that callbacks come from the CCS.
 
-```json
-{"url": "<callback url>", "auth": ... }
+```note
+TODO: What about using sane defaults for `endpoints`, `contest_ids` and `active` when adding a new webhook?
```
niemela, line 1601: The only possible (?) sane default would be empty. That's not a great default for an online setting though (but it works pretty great at an onsite event.)

Maybe `contest_ids` should be required? (At least if `/webhooks` is at the top level (see earlier comment).

Contest_API.md
```diff
@@ -1571,65 +1571,74 @@ If the client fails to respond to multiple requests over a period of
...
-
-` POST https://myurl`
-
-Payload:
-
-```json
-{"contest_id":"finals","endpoint":"teams","id":"11","data":{"id":"11","icpc_id":"201433","name":"The Shanghai Tigers","organization_id":"inst123","group_id":"asia"}}
-```
+When the CCS wants to send out a callback, it will check all active webhooks, filter them on applicable endpoint and contest ID and perform a `POST` to the URL.
+The CCS will add a header to this request called `Webhook-Token` which contains the token as supplied when creating the webhook.
```
niemela, line 1639: Is this what others do? Do we have some links to examples?

Contest_API.md
```diff
@@ -1571,65 +1571,74 @@ If the client fails to respond to multiple requests over a period of
...
-The following endpoint is associated with the webhook:
+The following endpoints is associated with the webhook:
 
-| Endpoint   | Mime-type        | Required? | Source @WF | Description                            |
-| ---------- | ---------------- | --------- | ---------- | -------------------------------------- |
-| `/webhook` | application/json | yes       | CCS        | List or register for webhook callbacks |
+| Endpoint         | Mime-type        | Required? | Source @WF | Description                                                                                                          |
+| ---------------- | ---------------- | --------- | ---------- | -------------------------------------------------------------------------------------------------------------------- |
+| `/webhooks`      | application/json | yes       | CCS        | JSON array of all webhook callbacks with elements as defined in the table below. Also used to register new webhooks. |
+| `/webhooks/<id>` | application/json | yes       | CCS        | JSON object of a single webhook callback with elements as defined in the table below                                 |
```
niemela, line 1579: We could have `/contests/<id>/webhooks` (and `/contests/<id>/webhooks/<id>`) instead (or in addition).

Putting it under contests allows for saner defaults. Not putting it at the top level makes it harder to follow *all* contests, which might be useful at an onsite event, but is not very good at all in an online setting.

Contest_API.md
nickygerritsen: https://docs.gitlab.com/ee/user/project/integrations/webhooks.html#secret-token
https://developer.github.com/webhooks/event-payloads/
They both do a header.

Contest_API.md
nickygerritsen: I'd say same list yes. Let's specify this in a general place? Maybe separate from this PR?

Contest_API.md
eldering: Decided it should at least be at the global level `/contests/webhooks`.

Contest_API.md
nickygerritsen: I will re-add the list from the old wiki and add `/webhooks/<id>` to it.

Contest_API.md
nickygerritsen: Yes. I will make sure add it and if you specify contest_ids it will never give webhooks (since it is not under any contest ID).

nickygerritsen: Decision: we will *not* perform callbacks for webhook updates. If one wants to know if a webhook is disabled he needs to call `/webhooks/<id>` to find out.

Contest_API.md
nickygerritsen: We will use empty arrays as defaults for `contest_ids` and `endpoints`.

We will remove `active` for now from the webhooks.

Contest_API.md
nickygerritsen: We will use `Webhook-Token` since. We will not use the `X-` prefix since it has been [deprecated since 2012](https://tools.ietf.org/html/rfc6648)

pehrsoderman (approved):

niemela (approved):

PR #9, opened 2020-11-02
Fix Baylor URLs and switch to Contest API

deboer-tim:
Minor URL fix, plus switch references from event feed to Contest API. This is really just changing a couple references, we still need to talk about the broader Contest API / contest archive / file format use throughout this spec.


deboer-tim:
Cancelling, I see another PR that does the URL fixes.

PR #14, opened 2020-12-28
Add support for POSTing submissions.

nickygerritsen:
As discussed during the last call, I made a first attempt at adding the `POST .../submissions` endpoint.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1062,9 +1062,12 @@ To add a submission one can use the `POST` method on the submissions endpoint.
 The `POST` must include a valid JSON object with the same attributes the submission
 endpoint returns with a `GET` request with the following exceptions:
 
+* The attributes `id`, `team_id`, `time`, and `contest_time` are
+  optional. However, depending on the use case (see below) the server
+  may require attributes to either be absent or present, and should
+  respond with a 400 error code in such cases.
 * Since `files` only supports `application/zip`, providing the `mime` field is
   optional.
-* `contest_time` must not be provided.
 * `reaction` may be provided but a CCS does not have to honour it.
 * The `time` attribute is optional. If not provided (or `null`) it will default
   to the current time as determined by the server.
@@ -1093,6 +1096,58 @@ The request must fail with a 400 error code if any of the following happens:
 The response will be the ID of the newly added submission.
 
 Performing a `POST` by any other roles than `admin` and `team` are not supported.
+
+#### Use cases for POSTing submissions
+
+The `POST submissions` endpoint can be used for a variety of reasons,
+and depending on the use case, the server might require different
+fields to be present. A number of common scenarios are described here
+for informational purposes only.
+
+##### Team client submitting to CCS
+
+The most obvious and probably most common case is where a team
+directly submits to the CCS, e.g. with a command-line submit client.
+
+In this case the client has the `team` role and a specific `team_id`
+already associated with it. The attributes `id`, `team_id`, `time`,
+and `contest_time` should not be specified; the server will
+determine these attributes and should reject submissions specifying
+them, or may ignore a `team_id` that is identical to the one of the
+current team.
+
+##### A proxy server forwarding to a CCS
+
+A proxy server may receive submissions from team clients (like above)
+and forward these to a CCS. This might be useful, for example, in a
+multi-site contest setup, where each site runs a proxy that would
+still be reachable if connectivity with the central CCS is lost, or
+where the proxy forwards the submission to multiple CCS's that run in
+parallel (like the shadowing setup at the ICPC World Finals).
+
+In such a scenario, the proxy server would timestamp the submissions
+and authenticate the submitting team, and then forward the submission
+to the upstream CCS using the `admin` role. The proxy would provide
+`team_id` and `time` attributes and the CCS should then accept and use
+these.
+
+To allow the proxy to return a submission `id` during connectivity
+loss, each site could be assigned a unique prefix such that the proxy
+server can itself generate unique `id`s and then submit to the central
+CCS with the `id` attribute included. The central CCS should then
+accept and use that `id` attribute.
+
+##### Further potential extensions
+
+To allow for any further use cases, the specification is deliberately
+flexible in how the server can handle optional attributes.
+
+* The `contest_time` attribute should normally not be specified when
+  `time` is already specified as it can be calculated from `time` and
+  the wallclock time is unambiguously defined without reference to
+  contest start time. However, in a case where one would want to
+  support a multi-site contest where the sites run out of sync, the
+  use of `contest_time` might be considered.
 
 #### Example
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1072,15 +1072,21 @@ endpoint returns with a `GET` request with the following exceptions:
   the ID of the team associated with the request and `time` will always use
   the current time as determined by the server. The CCS will determine an `id`.
 * If an `id` is supplied, the client should make sure it is unique, i.e. not used
-  yet on the CCS.
+  yet on the CCS. The client should normally not supply `id`, but let it be determined
+  by the server. However, for example in a setup with a central CCS with satellite sites
+  where teams submit to a proxy CCS that forwards to the central CCS, this might be
+  useful to make sure that the proxy CCS can accept submissions even when the connection
+  to the central CCS is down. The proxy can then forward these submissions later, when
+  the connection is restored again.
 
 The request should fail with a 400 if any of the following happens:
 
-* A required attribute is msising.
+* A required attribute is missing.
 * The supplied problem, team or language can not be found.
 * An entrypoint is required for the given language, but not supplied.
 * Something is wrong with the submission file. For example it contains too many
   files, it is too big, etc.
+* The provided `id` already exists.
 
 The response will be the ID of the newly added submission.
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1108,7 +1108,6 @@ Request data:
    "team_id": "123",
    "time": "2014-06-25T11:22:05.034+01",
    "entry_point": "Main",
-   "start_time": "2014-06-25T10:00:00+01",
    "files": [{"data": "<base64 string>"}]
 }
 ```
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1062,19 +1062,19 @@ To add submissions one can use the `POST` method on the submissions endpoint.
 The `POST` must include a valid JSON object with the same attributes the submission
 endpoint returns with a `GET` request with the following exceptions:
 
-* Any provided `id`, `reaction` and `contest_time` attributes are ignored.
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1058,6 +1056,37 @@ clients with `admin` or `analyst` role. The `reaction` attribute
...
+| ------------- | ---------------- | --------- | --------- | -------------------------------------------------------------------------------------------------- |
+| language\_id  | ID               | yes       | no        | identifier of the [ language](#languages) to submit for                                            |
+| problem\_id   | ID               | yes       | no        | identifier of the [ problem](#problems) to submit for                                              |
+| team\_id      | ID               | yes       | no        | identifier of the [ team](#teams) that made the submission                                         |
+| time          | TIME             | no        | yes       | timestamp of when the submission was made                                                          |
+| entry\_point  | string           | no        | yes       | code entry point for specific languages                                                            |
+| files         | array of ARCHIVE | yes       | no        | submission files, contained at the root of the archive. Only allowed mime type is application/zip. |
```
RagnarGrootKoerkamp, line 1071: Why `array of ARCHIVE` instead of just `ARCHIVE`, since they are already zipped. What's the difference in behavior.

nickygerritsen: In GET it is also an array while only one item is allowed so it is for consistency.
I think it is an array because it is an array in other places. But I’m not sure and have already said I don’t like it being an array

eldering: It was a design choice that for consistency a file-type attribute always returns an `array of ARCHIVE`. We've debated that pretty extensively, but the details of the pro/con arguments might be lost. We could rethink that (and this time record the arguments). And indeed, I think for consistency sake we should have the `POST` should use the same structure as the `GET`.

RagnarGrootKoerkamp: OK, consistency is a fine argument. In that case just explicitly mention here that the array must have length 1.

nickygerritsen: Here and in the GET 🙂

Contest_API.md
```diff
@@ -1058,6 +1056,26 @@ clients with `admin` or `analyst` role. The `reaction` attribute
 is available to clients with `public` role only when the contest is
 not frozen.
 
+#### POST submissions
+
+To add submissions one can use the `POST` method on the submissions endpoint.
+The `POST` must include a valid JSON object with the same attributes the submission
+endpoint returns with a `GET` request with the following exceptions:
+
+* Any provided `id`, `reaction` and `contest_time` attributes are ignored.
```
eldering, line 1065: We might also want to explicitly disallow these, see also the wording about `id` in https://ccs-specs.icpc.io/contest_api#http-methods
But just saying "should not be provided, and are ignored" is fine with me.

Contest_API.md
```diff
@@ -1058,6 +1056,26 @@ clients with `admin` or `analyst` role. The `reaction` attribute
...
+* The `time` attribute is optional. If not provided (or `null`) it will default
+  to the current time as determined by the server.
+* Since `files` only supports `application/zip`, providing the `mime` field is
+  optional.
+* If the CCS supports a `team` role, the `team_id` and `time`
+  attributes will be ignored when using a this role. `team_id` will then use
```
eldering, line 1071: Drop "a" in "using a this role". And "`team_id` will then be set to".

Contest_API.md
```diff
@@ -1058,6 +1056,26 @@ clients with `admin` or `analyst` role. The `reaction` attribute
...
+  the ID of the team associated with the request and `time` will always use
+  the current time as determined by the server.
+
+The response will be the ID of the newly added submission.
+
+The `public` role can never add submissions.
```
eldering, line 1077: Maybe better to be more explicit and say that `POST`s by any other roles than `admin` and `team` are not supported.

eldering (commented):
A few nitpicks, but look fine to me.

nickygerritsen:
A point @deboer-tim raised:

If we want to create a 'proxy' we need it to be roughly similar in behavior, so we should maybe say what happens if you give invalid data.

I propose the following addition:

> The request should fail with a 400 if any of the following happens:
> 
> * A required attribute is msising.
> * The supplied problem, team or language can not be found.
> * An entrypoint is required for the given language, but not supplied.
> * Something is wrong with the submission file. For example it contains too many
>   files, it is too big, etc.

Additionally, what should we do with the returned submission ID? If a proxy is not in sync with a CCS, it can't determine the submission ID yet. Should it determine an ID itself? Then it would also need to map it. Or do we leave a proxy open to not returning anything, meaning it doesn't have to have the exact same response data as a CCS it backs?

deboer-tim:
I think we'll probably need some more discussion around the failure cases, but +1 to everything else. IMHO we should go ahead and merge what we've got.

Contest_API.md
```diff
@@ -1062,15 +1062,25 @@ To add submissions one can use the `POST` method on the submissions endpoint.
...
 * Since `files` only supports `application/zip`, providing the `mime` field is
   optional.
-* If the CCS supports a `team` role, the `team_id` and `time`
+* If the CCS supports a `team` role, the `team_id`, `time` and `id`
   attributes will be ignored when using this role. `team_id` will then be set to
   the ID of the team associated with the request and `time` will always use
-  the current time as determined by the server.
+  the current time as determined by the server. The CCS will determine an `id`.
+* If an `id` is supplied, the client should make sure it is unique, i.e. not used
+  yet on the CCS.
```
eldering, line 1075: I'd add here something along the lines of:
The client should normally not supply `id`, but let it be determined by the server. However, for example in a setup with a central CCS with satellite sites where teams submit to a proxy CCS and that forwards to the central CCS, this might be useful to make sure that the proxy CCS can accept submissions even when the connection to the central CCS is down. Then the proxy can forward these submissions later, when the connection is restored again.

Contest_API.md
```diff
@@ -1062,15 +1062,25 @@ To add submissions one can use the `POST` method on the submissions endpoint.
...
+
+The request should fail with a 400 if any of the following happens:
+
+* A required attribute is msising.
```
eldering, line 1079: Typo: `missing`

Contest_API.md
```diff
@@ -1062,15 +1062,25 @@ To add submissions one can use the `POST` method on the submissions endpoint.
...
+* The supplied problem, team or language can not be found.
+* An entrypoint is required for the given language, but not supplied.
+* Something is wrong with the submission file. For example it contains too many
+  files, it is too big, etc.
```
eldering, line 1083: Add:
* The provided `id` already exists.

deboer-tim (commented):
Some additional thoughts as I'm trying to implement this:

- I think it's odd (and potentially confusing!) that a team role can include a random id, team_id, and time. I think it's cleaner if the submission will be rejected if a team role tries to specify any of these.
- Likewise, I think including contest_time should cause the call to fail.
- The time and id comments don't explicitly mention the admin role, and I think they should. e.g. 'If the server supports an admin role, team_id must be specified, id and time are optional...' then use the existing text to explain why you might (or might not) want to include these.
- If we ever had the situation where a local proxy was collecting submissions while the CCS wasn't accessible, reaction videos would also have to be stored locally, and then either forwarded with the submissions or updated afterward. I'd prefer to separate this from contest_time and put it in the admin section, maybe something like 'reaction can only be included by admin role and may not be supported'.

To sum it up, my proposal is to format it like this:
- mime type is optional
- contest_time is never allowed
- as a team role the id, team_id, and time aren't allowed
- as an admin role the team_id is required, id and time are optional (with explanatory text), reaction is optional and may not be honoured

eldering:
>     * If we ever had the situation where a local proxy was collecting submissions while the CCS wasn't accessible, reaction videos would also have to be stored locally, and then either forwarded with the submissions or updated afterward. I'd prefer to separate this from contest_time and put it in the admin section, maybe something like 'reaction can only be included by admin role and may not be supported'.

Maybe more of a side remark, but reaction videos might also be available only a bit after the submission itself, so in that case we'd need a PUT or PATCH.

deboer-tim:
> Maybe more of a side remark, but reaction videos might also be available only a bit after the submission itself, so in that case we'd need a PUT or PATCH.

Agreed. I'm ok handling that part later or as CDS-specific extension if necessary, I just didn't want to make it prohibited to include it in the case it is available.

nickygerritsen:
> Some additional thoughts as I'm trying to implement this:
> 
> * I think it's odd (and potentially confusing!) that a team role can include a random id, team_id, and time. I think it's cleaner if the submission will be rejected if a team role tries to specify any of these.
> * Likewise, I think including contest_time should cause the call to fail.
> * The time and id comments don't explicitly mention the admin role, and I think they should. e.g. 'If the server supports an admin role, team_id must be specified, id and time are optional...' then use the existing text to explain why you might (or might not) want to include these.
> * If we ever had the situation where a local proxy was collecting submissions while the CCS wasn't accessible, reaction videos would also have to be stored locally, and then either forwarded with the submissions or updated afterward. I'd prefer to separate this from contest_time and put it in the admin section, maybe something like 'reaction can only be included by admin role and may not be supported'.
> 
> To sum it up, my proposal is to format it like this:
> 
> * mime type is optional
> * contest_time is never allowed
> * as a team role the id, team_id, and time aren't allowed
> * as an admin role the team_id is required, id and time are optional (with explanatory text), reaction is optional and may not be honoured

I’m fine with these changes if others are. I also agree to discuss reaction separately

deboer-tim:
LGTM. I reviewed earlier, but adding a comment to make it clear I think we should go ahead and merge - any potential remaining issues are likely to be minor and will be easier to find and discuss on master.

nickygerritsen:
Agreed with Tim, looks good.

deboer-tim (approved):

Contest_API.md
```diff
@@ -1058,6 +1056,99 @@ clients with `admin` or `analyst` role. The `reaction` attribute
...
+To add a submission one can use the `POST` method on the submissions endpoint.
+The `POST` must include a valid JSON object with the same attributes the submission
+endpoint returns with a `GET` request with the following exceptions:
+
+* The attributes `id`, `team_id`, `time`, and `contest_time` are
+  optional. However, depending on the use case (see below) the server
+  may require attributes to either be absent or present, and should
+  respond with a 400 error code in such cases.
+* Since `files` only supports `application/zip`, providing the `mime` field is
+  optional.
```
niemela, line 1070: Is the server required to reject a POST that includes a mime that is not `application/zip`?

niemela (approved):
Might need some more iteration, but technically everything is currently marked as a "development draft", so we could/should go ahead and merge.

Contest_API.md
deboer-tim: As written the server isn't required to even look at the mime type, but would fail and reject later if the actual content wasn't a zip. IMHO this probably should be one of the failure requirements, but I would go ahead and merge and then discuss/PR this.

eldering: See #18 for this.

PR #15, opened 2021-02-24
Initial clarification post support

deboer-tim:
Allowing teams and admins to post clarifications to the contest

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1209,7 +1209,38 @@ Note that at least one of `from_team_id` and `to_team_id` has to be
 
 Clients with the `public` role can only view clarifications replies
 from the jury to all teams, that is, messages where both `from_team_id`
-and `to_team_id` are `null`.
+and `to_team_id` are `null`. Clients with the `team` role can only view
+their own clarifications (sent or received) and public clarifications.
+
+#### POST clarifications
+
+To add a clarification one can use the `POST` method on the clarifications endpoint.
+The `POST` must include a valid JSON object with the same attributes the clarification
+endpoint returns with a `GET` request with the following exceptions:
+
+* If the server supports a `team` role, `id`, `to_team_id`, `reply_to_id`, `time`, and
+  `contest_time` must not be provided. `from_team_id` may be provided but then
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1209,7 +1209,38 @@ Note that at least one of `from_team_id` and `to_team_id` has to be
...
+and `to_team_id` are `null`. Clients with the `team` role can only view
+their own clarifications (sent or received) and public clarifications.
+
+#### POST clarifications
+
+To add a clarification one can use the `POST` method on the clarifications endpoint.
+The `POST` must include a valid JSON object with the same attributes the clarification
+endpoint returns with a `GET` request with the following exceptions:
+
+* If the server supports a `team` role, `id`, `to_team_id`, `reply_to_id`, `time`, and
```
eldering, line 1221: I'd replace `If the server supports` with `When submitting using`, or at least add words along these lines. Now it is only implicit that this is when submitting as team role.

Contest_API.md
```diff
@@ -1209,7 +1209,38 @@ Note that at least one of `from_team_id` and `to_team_id` has to be
...
+  `contest_time` must not be provided. `from_team_id` may be provided but then
+  must match the ID of the team associated with the request. The server will determine
+  an `id` and the current `time` and `contest_time`.
```
eldering, line 1224: Do we also want to add the constraint for teams that `reply_to_id` can only be set to a clarification that is visible to that team?

Contest_API.md
deboer-tim: Agreed, will do in following commit.

Contest_API.md
deboer-tim: I had originally misread the reply_to and thought it could only be used by jury. Looking at it again I guess a team could reply to a reply, or reply to a broadcast? If so I agree with your comment, but I'd also need to remove it from the first sentence above that says it must not be provided.

nickygerritsen (approved):

niemela (approved):
Definitely good enough as the first draft :).

PR #23, opened 2021-03-29
Move changes from CAF into Contest API

niemela:
Include the changes done in the Contest Archive Format into the API. This removed most differences. Remaining are:

- Binary file references are not required to work when saved to disk, rather binary files are stored in a predetermined location.
- `countdown_pause_time` is not in the CAF.
- Wherever `decimal` is used in the API, the CAF uses `number`, which is a built-in.
- CAF has not been updated for webhooks (but technically, neither has the API at the time of writing this).

We probably want to fix the last two...?

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -739,7 +739,7 @@ JSON elements of problem objects:
 | Name              | Type    | Required? | Nullable? | Source @WF | Description                                                                                                                                                       |
 | ----------------- | ------- | --------- | --------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | id                | ID      | yes       | no        | CCS        | identifier of the problem, at the WFs the directory name of the problem archive                                                                                   |
-| uuid              | string  | no        | yes       | not used   | UUID of the problem, as defiend in the problem package. |
+| uuid              | string  | no        | yes       | not used   | UUID of the problem, as defined in the problem package. |
 | label             | string  | yes       | no        | CCS        | label of the problem on the scoreboard, typically a single capitalized letter                                                                                     |
 | name              | string  | yes       | no        | CCS        | name of the problem                                                                                                                                               |
 | ordinal           | ORDINAL | yes       | no        | CCS        | ordering of problems on the scoreboard                                                                                                                            |
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -640,6 +640,8 @@ JSON elements of Command objects:
 | version         | string | no       | Expected output from running the version-command.                                      |
 | version-command | string | no       | Command to run to get the version. Defaults to `<command> --version` if not specified. |
 
+The compiler and runner elements are intended for informational purposes. It is not expected that systems will synchronize compiler and runner settings via this interface.
+
 #### Access restrictions at WF
 
 No access restrictions apply to a GET on this endpoint.
@@ -737,7 +739,7 @@ JSON elements of problem objects:
 | Name              | Type    | Required? | Nullable? | Source @WF | Description                                                                                                                                                       |
 | ----------------- | ------- | --------- | --------- | ---------- | ----------------------------------------------------------------------------------------------------------------------------------------------------------------- |
 | id                | ID      | yes       | no        | CCS        | identifier of the problem, at the WFs the directory name of the problem archive                                                                                   |
-| uuid              | string  | no        | yes       | not used   | UUID of the problem. |
+| uuid              | string  | no        | yes       | not used   | UUID of the problem, as defiend in the problem package. |
 | label             | string  | yes       | no        | CCS        | label of the problem on the scoreboard, typically a single capitalized letter                                                                                     |
 | name              | string  | yes       | no        | CCS        | name of the problem                                                                                                                                               |
 | ordinal           | ORDINAL | yes       | no        | CCS        | ordering of problems on the scoreboard                                                                                                                            |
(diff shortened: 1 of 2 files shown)


eldering (dismissed):

Contest_API.md
```diff
@@ -621,14 +621,27 @@ The following endpoints are associated with languages:
...
-| id                   | ID              | yes       | no        | CCS        | identifier of the language from table below                                                                             |
-| name                 | string          | yes       | no        | CCS        | name of the language (might not match table below, e.g. if localized)                                                   |
+| Name                 | Type   | Required? | Nullable? | Source @WF | Description                                                           |
+| -------------------- | ------ | --------- | --------- | ---------- | --------------------------------------------------------------------- |
+| id                   | ID     | yes       | no        | CCS        | identifier of the language from table below                           |
+| name                 | string | yes       | no        | CCS        | name of the language (might not match table below, e.g. if localized) |
+| compiler             | Command object | no | yes      | CCS        | Command used for compiling submissions. |
```
nickygerritsen, line 628: I get why you used `Command Object` as type here, since you don't want to repeat yourself. However, this is the first time we link to a type that is only defined below (right?). I'm wondering if this is clear enough.

nickygerritsen (commented):
> Wherever decimal is used in the API, the CAF uses number, which is a built-in.

We should definitely fix this yes.

> CAF has not been updated for webhooks (but technically, neither has the API at the time of writing this).

Maybe not do this yet? Any of us still need to implement webhooks to see if our idea actually works.

Contest_API.md
niemela: Well the list at the top (under "Archive contents") refers to all the types, and they are all (obviously) defined below. That said, up there they are all linked, we should do the same here.

nickygerritsen:
> > > Wherever decimal is used in the API, the CAF uses number, which is a built-in.
> > 
> > 
> > We should definitely fix this yes.
> 
> Which way? I would prefer updating the API to use simple built-ins.
> 
I'm fine with that. Does indeed make sense to use built ins. However we do specify we only allow 3 decimals in the API spec. Do we want to keep that?

niemela:
> > [...] I would prefer updating the API to use simple built-ins.
> 
> I'm fine with that. Does indeed make sense to use built ins. However we do specify we only allow 3 decimals in the API spec. Do we want to keep that?

I would prefer to drop that, t.b.h.

Contest_API.md
niemela: A right, sorry. I was indeed referring to the CAF spec. In the API spec we always refer to object definitions below, just not by name. E.g. for the languages endpoint(s) the description is "JSON array of all languages with elements as defined in the table below". If the endpoint table had a type column (and maybe it should), it would say "Language object" and "Array of Language objects" respectively.

In the API spec the definition of the Command objects is literally *directly* below the (very short) table where it is referred to. I think that is close enough for a forward reference to work without further explanation. Also, there is no actual header above the Command object, so we can't actually link to it (although, we could of course change that).

In summary, I suggest no change regarding this. :)

nickygerritsen: OK fair point

niemela:
> Why do we even have it? It was from before my time

It's used for `time_limit`, `max_run_time` and `run_time`. The idea was to force a certain (maximal as well as minimal) resolution. I think it was a mistake to do that by defining a special type, we should just add it as a "should" in the descriptions.

Contest_API.md
```diff
@@ -621,14 +621,27 @@ The following endpoints are associated with languages:
...
 | entry_point_required | boolean         | yes       | no        | CCS        | whether the language requires an entry point                                                                            |
 | entry_point_name     | string          | depends   | yes       | CCS        | the name of the type of entry point, such as "Main class" or "Main file"). Required iff entry_point_required is present |
 | extensions           | array of string | yes       | no        | CCS        | file extensions for the language                                                                                        |
 
+JSON elements of Command objects:
+
+| Name            | Type   | Required | Description |
+| :-------------- | :----- | :------- | :---------- |
+| command         | string | yes      | Command to run.                                              |
+| args            | string | no       | Argument list for command. {files} denotes where to include the file list. |
```
eldering, line 639: Nitpick: put `{files}` in backquotes.

niemela: Done.

eldering (dismissed):

pehrsoderman (dismissed):

niemela:
Will rebase on master and merge.

Contest_API.md
```diff
@@ -621,13 +621,39 @@ The following endpoints are associated with languages:
...
+| version-command | string | no       | Command to run to get the version. Defaults to `<command> --version` if not specified. |
+
+The compiler and runner elements are intended for informational purposes. It is not expected that systems will synchronize compiler and runner settings via this interface.
+
+JSON elements of Command objects:
+
+| Name            | Type   | Required | Description |
+| :-------------- | :----- | :------- | :---------- |
+| command         | string | yes      | Command to run.                                              |
+| args            | string | no       | Argument list for command. `{files}` denotes where to include the file list. |
```
meisterT, line 652: Late to the party, but we need other vars similar `{files}`, e.g. for derivatives of the entry point (think main class in Java and Kotlin).

meisterT: Also for Java the heap size is dependent on the problem's memory limit, so we would need a way to express that.

deboer-tim: This issue was merged a year ago, so I'm worried that comments here might be forgotten. I'd open up an issue, submit a PR with a suggestion (even if not complete, it'll generate discussion) or join an upcoming call.

nickygerritsen: I promised Tobi to put this in the agenda, which I just did.

niemela: > Late to the party, but we need other vars similar `{files}`,

Yes. agreed.

PR #27, opened 2021-04-26
Drop system.yaml and move clarification related fields to hardcoded lists

nickygerritsen:
I'm struggling a bit with if and how we should word that the non-problem categories don't HAVE to be assignable by teams but only admins (as long as the teams can choose something like 'no specific problem).

Should we? And if so, how?

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -274,20 +274,23 @@ on the contest network.
 
 ### Predefined Clarification Answers
 
-The CCS must be able to configure predefined answers to clarification
-requests (e.g., "No response, read problem statement." and "This will be
-answered after the practice session.") from the [ contest configuration
-file](#contestyaml), so that judges can choose to reply to a
-clarification request by selecting a predefined answer rather than being
-required to enter a specific answer. One of the predefined answers must
-be the "default" answer.
+The CCS must support predefined answers to clarification requests, so
+that judges can choose to reply to a clarification request by selecting
+a predefined answer rather than being required to enter a specific
+answer. At least the following predefined answers must be supported:
+
+* No comment, read problem statement.
+* This will be answered during the answers to questions session.
 
 ### Clarification Categories
 
-The CCS must be able to configure "categories" to which clarification
-requests can be assigned from the [ contest configuration
-file](#contestyaml). A request belongs to exactly one
-category. Examples of categories are "General", "SysOps", "Operations".
+The CCS must support "categories" to which clarification
+requests can be assigned. A request belongs to exactly one
+category. At least the following categories need to be supported:
+
+* General
+* SysOps
+* Operations
 
 In addition, the CCS must construct one category per problem, i.e.,
 categories named e.g. "Problem A", "Problem B", etc. for each problem.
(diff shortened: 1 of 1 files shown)


Contest_Control_System_Requirements.md
```diff
@@ -274,20 +274,23 @@ on the contest network.
...
-clarification request by selecting a predefined answer rather than being
-required to enter a specific answer. One of the predefined answers must
-be the "default" answer.
+The CCS must support predefined answers to clarification requests, so
+that judges can choose to reply to a clarification request by selecting
+a predefined answer rather than being required to enter a specific
+answer. At least the following predefined answers must be supported:
+
+* No comment, read problem statement.
+* This will be answered during the answers to questions session.
```
niemela, line 283: I would suggest dropping the second and only requiring the first.

Contest_Control_System_Requirements.md
```diff
@@ -1222,66 +1225,6 @@ scoreboard-freeze-length: 1:00:00
...
-
-A YAML file consisting of a mapping with the following keys:
-
-| Key             | Description                                                                                      |
-| --------------- | ------------------------------------------------------------------------------------------------ |
-| default-clars   | Sequence of pre-defined clarification answers. The first is the default and will be pre-selected |
-| clar-categories | Sequence of categories for clarifications.                                                       |
-| languages       | Sequence of mappings with keys as defined below                                                  |
-
-##### languages
```
niemela, line 1235: We're not actually dropping the language information, but this should instead be read from the CAF. That said, this section describes how languages are listed in system.yaml, so that should be removed, but maybe we should exchange it for a reference to the same information in the CAF at the same time?

niemela (changes requested):

Contest_Control_System_Requirements.md
```diff
@@ -274,20 +274,23 @@ on the contest network.
 
 ### Predefined Clarification Answers
 
-The CCS must be able to configure predefined answers to clarification
-requests (e.g., "No response, read problem statement." and "This will be
-answered after the practice session.") from the [ contest configuration
-file](#contestyaml), so that judges can choose to reply to a
```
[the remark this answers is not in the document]
nickygerritsen: Ha, yes.

Contest_Control_System_Requirements.md
nickygerritsen: Why? Isn't the second used more often than the first? I think it is the most used response during dress rehearsal / practice session and with that perhaps also the most used overall. What would dropping it do?

Contest_Control_System_Requirements.md
nickygerritsen: *should* or *could*? I mean, DOMjudge never read the language data from system.yaml and might not read the same data from the CAF, but we still work correctly.
We already link to the ICPC programming environment in [this section](https://ccs-specs.icpc.io/ccs_system_requirements#supported-languages), which list languages and compile options. We could add a sentence there that reads something like:

> The language options can also be read from the [Contest Archive Format](https://ccs-specs.icpc.io/contest_archive_format#language-object)

Contest_Control_System_Requirements.md
niemela: > What would dropping it do?

Shorten the specification?

Maybe you're right...

nickygerritsen: I like this last statement of yours

Contest_Control_System_Requirements.md
niemela: > _should_ or _could_? 

Not sure... whatever it is now?

> I mean, DOMjudge never read the language data from system.yaml and might not read the same data from the CAF, but we still work correctly.

Yeah... I've always been of the opinion that humans can be part of the configuration system. I.e. if you look at the system.yaml file and then configure languages "manually", then you're "using" system.yaml. Same with the archive. It would be *very* nice if on top of that we could automatically *verify* that configuration has been done correctly, and including the compiler/interpreter information in the language endpoint is a big step in that direction.

> We could add a sentence there that reads something like:
> 
> > The language options can also be read from the [Contest Archive Format](https://ccs-specs.icpc.io/contest_archive_format#language-object)

I think we need to update the [Importing Contest Configuration section](http://localhost:4000/ccs_system_requirements#importing-contest-configuration). That does not *have* to be part of this PR though.

niemela (approved):

PR #28, opened 2021-05-10
Add support for filtering on ID's for collections.

nickygerritsen:


Contest_API.md
```diff
@@ -371,6 +371,15 @@ attributes that are:
...
     `null` value or because it was not implemented.
 
+### Filtering
+
+Endpoints that return a JSON array must allow filtering on any
+attribute with type ID (except the `id` field) by passing it as a
+query argument. For example, clarifications can be filtered on the
+recipient by passing `to_team_id=X`. To filter on a `null` value,
+pass an empty string, i.e. `to_team_id=`. It must be possible to filter
+on mulitple fields simultaneously.
```
eldering, line 381: Typo: "multiple".
Maybe a bit more explicit: "It must be possible to filter on multiple different fields simultaneously, with the meaning that all conditions must be met (they are logically `AND`ed)."

Contest_API.md
```diff
@@ -371,6 +371,15 @@ attributes that are:
     In the latter case that can either be because it was a left out
     `null` value or because it was not implemented.
 
+### Filtering
+
+Endpoints that return a JSON array must allow filtering on any
+attribute with type ID (except the `id` field) by passing it as a
```
eldering, line 377: What about attributes that are an array of IDs, like `group_ids` in the teams endpoint and `team_ids` in awards?
I'd argue to start simple and not support these yet.

nickygerritsen: Thats what I only mention ID. I could explicitly state that filtering on array of ID does not have to be supported

nickygerritsen: Added a note about array of ID

Contest_API.md
```diff
@@ -371,6 +371,18 @@ attributes that are:
...
+query argument. For example, clarifications can be filtered on the
+recipient by passing `to_team_id=X`. To filter on a `null` value,
+pass an empty string, i.e. `to_team_id=`. It must be possible to
+filter on multiple different fields simultaneously, with the
+meaning that all conditions must be met (they are logically `AND`ed).
+Note that filtering on any other field, including fields with the type
+array of ID, does not have to be supported.
```
niemela, line 384: "array of ID" should be backticked.

niemela (changes requested):

Contest_API.md
nickygerritsen: Then `ID` above should be as well? It is also not backticked in the tables though

niemela: > Then `ID` above should be as well? 

Good point.

> It is also not backticked in the tables though

Yeah... You know what, let's not...

(And some point we should think through what exactly should be quoted how and then make sure we are consistent. But that's just editing, and can trivially be done later. Let's not block relevant updates on that.)

niemela (approved):

eldering (approved):

PR #29, opened 2021-05-11
Add requirement for JSON response body on error

deboer-tim:
Added requirement that failure messages include a JSON response body with code and message attributes with further information for the user.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -142,7 +142,14 @@ their expected behavior, if implemented.
 
 Standard [HTTP status
 codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
-returned to indicate success or failure.
+returned to indicate success or failure. When there is a failure, the
+response message body must include a JSON element that contains the attributes 'code' and 'message' with further information suitable for the user making the
+request, as per the following example:
+
+```json
+{"code":"400",
+ "message":"Teams cannot send clarifications to another team"}
+ ```
 
 ### Roles
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -142,12 +142,11 @@ their expected behavior, if implemented.
 
 Standard [HTTP status
 codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
-returned to indicate success or failure. When there is a failure, the
-response message body must include a JSON element that contains the attributes 'code' and 'message' with further information suitable for the user making the
-request, as per the following example:
+returned to indicate success or failure. When there is a failure (status codes 4xx or 5xx), the
+response message body must include a JSON element that contains the attributes 'code' (a number, identical to the HTTP status code returned) and 'message' (a string) with further information suitable for the user making the request, as per the following example:
 
 ```json
-{"code":"400",
+{"code":400,
  "message":"Teams cannot send clarifications to another team"}
  ```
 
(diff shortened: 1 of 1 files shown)


nickygerritsen (dismissed):

Contest_API.md
```diff
@@ -142,7 +142,14 @@ their expected behavior, if implemented.
 
 Standard [HTTP status
 codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
-returned to indicate success or failure.
+returned to indicate success or failure. When there is a failure, the
+response message body must include a JSON element that contains the attributes 'code' and 'message' with further information suitable for the user making the
```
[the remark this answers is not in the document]
deboer-tim: All good points, fixed in second commit.

Contest_API.md
```diff
@@ -142,7 +142,13 @@ their expected behavior, if implemented.
 
 Standard [HTTP status
 codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
-returned to indicate success or failure.
+returned to indicate success or failure. When there is a failure (status codes 4xx or 5xx), the
+response message body must include a JSON element that contains the attributes 'code' (a number, identical to the HTTP status code returned) and 'message' (a string) with further information suitable for the user making the request, as per the following example:
```
niemela, line 146: There will (maybe?) not always be a good error message. I think it would be better for `message` to be optional.

This kind of setup is not uncommon. Do we know how other (good) APIs handle this?

Contest_API.md
```diff
@@ -142,7 +142,13 @@ their expected behavior, if implemented.
...
+
+```json
+{"code":400,
+ "message":"Teams cannot send clarifications to another team"}
```
niemela, line 150: This is not a *great* example, because the code is arguably wrong. I would say that trying to send a message to a team when authenticated as a team would be a `403 Forbidden`.

niemela (changes requested):

Contest_API.md
nickygerritsen: DOMjudge uses a package for its API and that uses exactly tje way Tim suggested, so it seems this way is at least used more

niemela: @nickygerritsen What package are you using?

Is it also treating `message` as a mandatory field?

nickygerritsen: I think it is in the default Symfony (a php framework) code but I will double check when on a bigger screen.

I think it always exposes a message, since it uses exceptions to generate them and exceptions have a message

deboer-tim: FWIW, I looked at a few REST APIs for background, most were either missing error handling or overly complicated. I did find this used in two other places. I don't remember one but the other is an OpenAPI/Swagger example: https://swagger.io/docs/specification/describing-responses/ (note that code is a string, maybe that's where I got it from :) ).

IMHO the message should be mandatory so that the client always has something it can show the user along with the code. The server can keep it generic/error code lookup if necessary, but in my experience there's almost always something more useful to say.

Contest_API.md
deboer-tim: Example updated to be a 403. I also noticed the clarification POST says you must return a "400 error code" so I changed this to a "4xx HTTP status code".

niemela (approved):

Contest_API.md
nickygerritsen: FYI the library we use and that does this by default is FOSRestBundle, where it has [this logic](https://github.com/FriendsOfSymfony/FOSRestBundle/blob/3.x/Serializer/Normalizer/FlattenExceptionHandler.php#L120-L121).
It always has a message and the code is an integer.

eldering (approved):

PR #32, opened 2021-05-11
Group scoreboard, types

deboer-tim:
Added the group scoreboard as discussed on the last call, the known group type "site", and further clarifications/explanation of existing rules.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -759,7 +759,8 @@ university teams, a group for corporate teams, and a group for ICPC-eligible tea
 belong to two or three of these.
 When there are different kinds of groups for different purposes (e.g. sites vs divisions), each
 group or set of groups should have a different type attribute
-(e.g. `"type":"site"` and `"type":"division"`).
+(e.g. `"type":"site"` and `"type":"division"`). It is recommended that each team is in no more
+than one group of a given type.
 
 Groups must exist for any combination of teams that must be ranked on a
 [group scoreboard](#group-scoreboard), which means groups may be created for combinations of
@@ -799,9 +800,9 @@ The list below contains standardized identifiers for known group
 types. These identifiers should be used when the purpose
 of a group matches.
 
-| Type  | Name                | Description
-| ----- | ------------------- | ---------------------------------
-| site  | Site                | A physical location where teams are competing, e.g. the "Hawaii site"
+| Type  | Description
+| ----- | ---------------------------------
+| site  | A physical location where teams are competing, e.g. the "Hawaii site"
 
 #### Examples
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -765,7 +765,7 @@ Groups must exist for any combination of teams that must be ranked on a
 [group scoreboard](#group-scoreboard), which means groups may be created for combinations of
 other groups. For instance, if there is a requirement to show a scoreboard for teams in each of `D`
 divisions at every one of `S` sites, then in addition to the `D` + `S` groups there will also be
-`D`x`S` combined/product groups. It is strongly recommended that these groups have a type like
+`D`x`S` combined/product groups. It is recommended that these groups have a type like
 `"type":"<group1>-<group2>"`, e.g. `"type":"site-division"`.
 
 The following endpoints are associated with groups:
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -751,8 +751,20 @@ Returned data:
 
 ### Groups
 
-Grouping of teams. At the World Finals these are the super regions, at
-regionals these are often different sites.
+Grouping of teams. At the World Finals these are the super regions; at other contests these
+may be the different sites, divisions, or types of contestants.
+
+Teams may belong to multiple groups. For instance, there may be a group for each site, a group for
+university teams, a group for corporate teams, and a group for ICPC-eligible teams. Teams could
+belong to two or three of these.
+When there are different kinds of groups for different purposes (e.g. sites vs divisions), each
+group or set of groups should have a different type attribute
+(e.g. `"type":"site"` and `"type":"division"`).
+
+Groups must exist for any combination of teams that should be ranked on a
+[group scoreboard](#group-scoreboard), which means groups may be created for combinations.
+For instance, if there's a requirement to show a scoreboard for teams in each of `D`
+divisions at every one of `S` sites, then there will also be `D`x`S` groups with a type like `"type":"site-division"`.
 
 The following endpoints are associated with groups:
 
@@ -773,11 +785,21 @@ JSON elements of group objects:
 | icpc\_id | string  | no        | yes       | CCS        | external identifier from ICPC CMS                                        |
 | name     | string  | yes       | no        | CCS        | name of the group                                                        |
 | type     | string  | no        | yes       | CCS        | type of this group                                                       |
-| hidden   | boolean | no        | yes       | CCS        | if group should be hidden from scoreboard. Defaults to false if missing. |
+| hidden   | boolean | no        | yes       | CCS        | if the group should be excluded from the [scoreboard](#scoreboard). Defaults to false if missing. |
 
 #### Access restrictions at WF
 
 No access restrictions apply to a GET on this endpoint.
+
+#### Known group types
+
+The list below contains standardized identifiers for known group
+types. These identifiers should be used when the purpose
+of a group matches.
+
+| Type  | Name                | Description
+| ----- | ------------------- | ---------------------------------
+| site  | Site                | A physical location where teams are competing, e.g. the "Hawaii site"
 
 #### Examples
 
@@ -1537,7 +1559,7 @@ Returned data:
 
 ### Scoreboard
 
-Scoreboard of the contest.
+Scoreboard for the contest.
 
 Since this is generated data, only the `GET` method is allowed here,
 irrespective of role.
@@ -1568,6 +1590,18 @@ returned. The request will fail with a 400 error if the id is invalid.
 A suggested efficient server-side implementation to provide this, is to
 store with each event that changes the scoreboard, the new team
 scoreboard row.
+
+##### Group scoreboard
+
+By passing `group_id` with a valid group ID a scoreboard can be requested for the teams in a particular group:
+
+`/scoreboard?group_id=site1`
+
+Each group scoreboard is ranked independently and contains only the teams that belong to the
+specified group. If a client wants to know 'local' vs 'global' rank it can query both the group and primary scoreboards.
+
+A 4xx error will be returned if the group id is not valid. Groups that are hidden to the role making
+the request are not valid.
 
 #### Scoreboard format
 
@@ -1613,9 +1647,10 @@ Each problem object within the scoreboard consists of:
 
 #### Access restrictions at WF
 
-For clients with the `public` role, the scoreboard will not include
-information from judgements of submissions received after the scoreboard
-has been frozen until it has been thawed.
+For clients with the `public` or `team` role, the scoreboard will not
+include teams that are in any group that is hidden, nor information
+from judgements of submissions received after the scoreboard has been
+frozen until it has been thawed.
 
 #### Example
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -759,8 +759,7 @@ university teams, a group for corporate teams, and a group for ICPC-eligible tea
 belong to two or three of these.
 When there are different kinds of groups for different purposes (e.g. sites vs divisions), each
 group or set of groups should have a different type attribute
-(e.g. `"type":"site"` and `"type":"division"`). It is recommended that each team is in no more
-than one group of a given type.
+(e.g. `"type":"site"` and `"type":"division"`).
 
 Groups must exist for any combination of teams that must be ranked on a
 [group scoreboard](#group-scoreboard), which means groups may be created for combinations of
@@ -802,7 +801,7 @@ of a group matches.
 
 | Type  | Description
 | ----- | ---------------------------------
-| site  | A physical location where teams are competing, e.g. the "Hawaii site"
+| site  | A physical location where teams are competing, e.g. the "Hawaii site". Teams generally should not be in more than one group of this type.
 
 #### Examples
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -761,10 +761,12 @@ When there are different kinds of groups for different purposes (e.g. sites vs d
 group or set of groups should have a different type attribute
 (e.g. `"type":"site"` and `"type":"division"`).
 
-Groups must exist for any combination of teams that should be ranked on a
-[group scoreboard](#group-scoreboard), which means groups may be created for combinations.
-For instance, if there's a requirement to show a scoreboard for teams in each of `D`
-divisions at every one of `S` sites, then there will also be `D`x`S` groups with a type like `"type":"site-division"`.
+Groups must exist for any combination of teams that must be ranked on a
+[group scoreboard](#group-scoreboard), which means groups may be created for combinations of
+other groups. For instance, if there is a requirement to show a scoreboard for teams in each of `D`
+divisions at every one of `S` sites, then in addition to the `D` + `S` groups there will also be
+`D`x`S` combined/product groups. It is strongly recommended that these groups have a type like
+`"type":"<group1>-<group2>"`, e.g. `"type":"site-division"`.
 
 The following endpoints are associated with groups:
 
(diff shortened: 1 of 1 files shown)


nickygerritsen (dismissed):

Contest_API.md
```diff
@@ -751,8 +751,20 @@ Returned data:
...
+may be the different sites, divisions, or types of contestants.
+
+Teams may belong to multiple groups. For instance, there may be a group for each site, a group for
+university teams, a group for corporate teams, and a group for ICPC-eligible teams. Teams could
+belong to two or three of these.
+When there are different kinds of groups for different purposes (e.g. sites vs divisions), each
+group or set of groups should have a different type attribute
+(e.g. `"type":"site"` and `"type":"division"`).
+
+Groups must exist for any combination of teams that should be ranked on a
```
niemela, line 764: Warning: nitpick.

It's not a **must** exist if it is only **should** be on a scoreboard. Best is probably to change both to "must"?

Contest_API.md
```diff
@@ -751,8 +751,20 @@ Returned data:
...
+[group scoreboard](#group-scoreboard), which means groups may be created for combinations.
+For instance, if there's a requirement to show a scoreboard for teams in each of `D`
+divisions at every one of `S` sites, then there will also be `D`x`S` groups with a type like `"type":"site-division"`.
```
niemela, line 767: I would suggest explicitly recommending the naming scheme `<group1>-<group2>` (i.e. "group names separated by dash") for product groups, rather than "just" recommending by example.

Contest_API.md
```diff
@@ -773,12 +785,22 @@ JSON elements of group objects:
 | icpc\_id | string  | no        | yes       | CCS        | external identifier from ICPC CMS                                        |
 | name     | string  | yes       | no        | CCS        | name of the group                                                        |
 | type     | string  | no        | yes       | CCS        | type of this group                                                       |
-| hidden   | boolean | no        | yes       | CCS        | if group should be hidden from scoreboard. Defaults to false if missing. |
+| hidden   | boolean | no        | yes       | CCS        | if the group should be excluded from the [scoreboard](#scoreboard). Defaults to false if missing. |
```
niemela, line 788: This is not really a "should" in the RFC 2119 sense. Maybe we should say "if the group is to be excluded..." instead?

Contest_API.md
```diff
@@ -773,12 +785,22 @@ JSON elements of group objects:
...
 
+#### Known group types
+
+The list below contains standardized identifiers for known group
+types. These identifiers should be used when the purpose
+of a group matches.
+
+| Type  | Name                | Description
+| ----- | ------------------- | ---------------------------------
+| site  | Site                | A physical location where teams are competing, e.g. the "Hawaii site"
```
niemela, line 802: Warning: Not actually a comment on this PR.

Wait, what? Is the ending pipe optional? They are always annoying. A world of possibilities opens up...

niemela (changes requested):

Contest_API.md

Contest_API.md
deboer-tim: Rewrote this paragraph a little, including fixing this in new commit.

Contest_API.md

Contest_API.md
deboer-tim: *Thank you* for ending my day with a laugh. Last night when I copied this from one of the other 'known groups' sections I was trying to figure out why the author took the time to put in all the end pipes. lol :)

niemela (approved):

Contest_API.md
```diff
@@ -751,8 +751,22 @@ Returned data:
...
-regionals these are often different sites.
+Grouping of teams. At the World Finals these are the super regions; at other contests these
+may be the different sites, divisions, or types of contestants.
+
+Teams may belong to multiple groups. For instance, there may be a group for each site, a group for
+university teams, a group for corporate teams, and a group for ICPC-eligible teams. Teams could
+belong to two or three of these.
+When there are different kinds of groups for different purposes (e.g. sites vs divisions), each
+group or set of groups should have a different type attribute
+(e.g. `"type":"site"` and `"type":"division"`).
```
eldering, line 762: I think we've discussed this before, but don't we want to require (or at least recommend) that for each group type, a team can at most (or exactly) be in one of those?

Contest_API.md
eldering: The way this table is written makes it sound like it's something that we explicitly refer to from other parts of the spec, but that's not the case, right? Also, the columns `Type` and `Name` make it unclear to me whether the site type should be put in as `site` or `Site`. I guess the former in lower case, so do we need the `Name` column?

Also, we should add at least `division` here, and maybe others we know of?

Contest_API.md
```diff
@@ -1613,9 +1649,10 @@ Each problem object within the scoreboard consists of:
 
 #### Access restrictions at WF
 
-For clients with the `public` role, the scoreboard will not include
-information from judgements of submissions received after the scoreboard
-has been frozen until it has been thawed.
+For clients with the `public` or `team` role, the scoreboard will not
+include teams that are in any group that is hidden, nor information
```
eldering, line 1653: Just a question: what if a team is part of both visible and hidden groups? I guess there are arguments both for making such a team visible or not. Was this discussed?
Is it is now, it's easy to make a team hidden, which is convenient e.g. for jury/admin associated teams. OTOH, it makes it difficult to assign some hidden groups that might only be visible to admins/jury.

Contest_API.md
deboer-tim: > The way this table is written makes it sound like it's something that we explicitly refer to from other parts of the spec, but that's not the case, right?

Correct. It's mostly just a hint for clients.

> Also, the columns `Type` and `Name` make it unclear to me whether the site type should be put in as `site` or `Site`. I guess the former in lower case, so do we need the `Name` column?

Yeah, I waffled on this before. Name column removed now.

> Also, we should add at least `division` here, and maybe others we know of?

On the call I had to argue that site should be listed because it has a useful meaning that would change defaults or behaviour for many tools (e.g. balloon printer, scoreboard for site, awards for site, etc). I'm not sure you can say the same for division, and while I don't think there's any harm in including I'd rather leave out to start and confirm at the next meeting.

Contest_API.md
deboer-tim: I added a recommendation.

Contest_API.md
deboer-tim: We didn't discuss this, and I wasn't initially planning on addressing it. When doing the rest I noticed that we didn't specify what to do if a team is in both non-hidden and hidden groups, nor whether a team should see their own solved runs in the scoreboard during freeze. I had a short chat with Nicky comparing what CDS and DOMjudge do (surprise, it wasn't consistent) and decided the above was the most consistent way to address this.

To do what you're suggesting I think we would need to separate whether a group is hidden from whether it hides teams on the scoreboard, which means a new (likely optional) attribute. If you think it's a case we should support then I think we should discuss on the next call.

Contest_API.md
niemela: > On the call I had to argue that site should be listed because it has a useful meaning that would change defaults or behaviour for many tools (e.g. balloon printer, scoreboard for site, awards for site, etc). 

You didn't have to argue *that* hard :).

> I'm not sure you can say the same for division, and while I don't think there's any harm in including I'd rather leave out to start and confirm at the next meeting.

Not only that, I don't think we actually have any known instances of division being used (or rather, known instances where division would have been used had the system followed the spec we are creating). Remember that the PacNW or SEUSA style divisions would **not** use a group of type division, rather they would be separate contests.

Well... actually, the NADC divisions would fit perfectly, but that is the first use, and not how we have typically thought about divisions. So, I'm not against adding division at some point, but we should at least have a good feeling for what that entails. If the only semantic meaning is "some sort of grouping", i.e. if it adds no additional semantic value on "being a group" and possibly "different types are different", then IMO we shouldn't add it.

Other kinds of groups that have been used in the past are "countries" (i.e. "Sweden", "Finland", "Norway", a.s.o.) or "subcontests" (i.e. "NCPC", "BAPC", "UKIEPC", a.s.o.). I don't think the former is a good choice since the concept of "country" has a lot of baggage, I would prefer to generalize it to something like "geographic area". The latter might be useful though?

Contest_API.md
niemela: > don't we want to require (or at least recommend) that for each group type, a team can at most (or exactly) be in one of those?

We should **absolutely** not require that for all group types. We could possibly add additional requirements for certain predefined types. E.g. it could make sense that if you are using groups of type "site" then every team has to be in exactly one site group.

> I added a recommendation.

I don't like that recommendation.

Contest_API.md
deboer-tim: 'Lightly convince', maybe? ;-)

I do agree with the rest. Site has a clear meaning, is being used, and can improve client behaviour, we should generally have similar reasons to add others.

Contest_API.md
deboer-tim: Ok, I removed it. I did move a waffly "generally should not" to the site description - IMHO maybe this should be stronger, but I'd rather get to a point where people are happy to merge and debate tweaks on a call.

Contest_API.md
niemela: There are two different concepts here. Hiding all the members of the group or hiding the group itself. The meaning of `hidden` in groups is (and has always been) the former. That said, the latter would definitely be more natural and make more sense. Since the meaning is "hide the members of the group", IMO the only reasonable interpretation is that you are hidden from the (main) scoreboard if you are a member of *any* hidden group, regardless of whether you are a member of any non-hidden groups.

All that said, I have never liked how we use `hidden`. I consider it a hack to allow for a commonly used feature (in DOMjudge) where you can have not only subset scoreboards (which is what groups allow for) but also superset  scoreboards (i.e. there are scoreboards that contain more teams then the default scoreboard, i.e. the hidden teams). 

I think a nicer way to support this could be to have a "default" group. At most one group may be the default group, and the teams in the default group is what you get if you query `/scoreboard` without a group_id argument (i.e. if `allteams` is the default group then `/scoreboard` and `/scoreboard?group_id=allteams` returns the exact same scoreboard). Furthermore, if no group is explicitly set as default, then the implicit group "all the teams" is default (otherwise you would almost always need to add an extra group of all teams just to make it the default).

Taking this even further (and maybe too far?). We do have contests where there is no scoreboard of all the teams. Specifically the NAQ and NADC. For these contests you *could* define an empty group to be the default group. This would cause `/scoreboard` to return an empty scoreboard, and you would *have* to specify a group_id to get anything useful. (It would kinda break the concept of local vs global rank referred ti in the group scoreboard section, but t.b.h. so does having any hidden team already.

Contest_API.md
eldering: > > don't we want to require (or at least recommend) that for each group type, a team can at most (or exactly) be in one of those?
> 
> We should **absolutely** not require that for all group types. We could possibly add additional requirements for certain predefined types. E.g. it could make sense that if you are using groups of type "site" then every team has to be in exactly one site group.

When does it make sense for a team to be part of multiple groups of the same group type? (Maybe I forgot a case we discussed before, but if anything I guess it would be useful to document here.)

niemela: Well, since groups and types could be anything, the answer very much depends on how you use them.

here are some cases from the top of my head:
* If there are groups for "has a graduate program" and "offers doctorates degrees", a school could be in one, both or neither.
* When you have filtering rules such as "ICPC eligible" as we do at NCPC, you end up with a group for all and a group for only eligible, but you typically don't have one for only non-eligible. Assuming you further filter on e.g. country (again, as we do at NCPC), then the country level ICPC and non-ICPC groups would (maybe?) be of the same type, but all eligible teams would be in both.
* If you have a country scoreboard but the semantics is not "country of the university" (maybe because the teams do not represent universities) but rather "nationality of the team members", then a team with team members of different nationalities would be in several of these groups.

The actual point though is that we have not in any way specified exactly how groups are to be used, so we definitely can't assume that the groups of a certain type has properties such as non-overlapping or completeness (i.e. every team is in one group of the type). Again, it would be a good idea to add these guarantees to predefined specific types when that makes sense. E.g. it *does* make sense to say that site is a proper partition of the teams.

eldering: Ok, but in your examples I don't see groups that could overlap *and* would naturally have the same type. So I think it's a reasonable requirement that a team can be part of _at most_ one group of each type. Thus a type is not yet a strict partition, but at least if you'd add a "other" group of that type it would be.

I think this requirement would be useful, for example for a tool that would want to display the "site" of a team. If you know that a team can be assigned at most one "site" group, then you can render that in a nicer way than if it could belong to multiple sites.

niemela: > Ok, but in your examples I don't see groups that could overlap _and_ would naturally have the same type.

The first and third are exactly that. In the third example we would have groups for every country (with participants in the contest). Every team with a Swedish team member would be in the "Sweden" group, every team with a Canadian team member would be in the "Canada" group. So, if me and Tim make a team, our team would be in both those groups. Are you saying that these country groups would not be of the same type?

> I think this requirement would be useful, for example for a tool that would want to display the "site" of a team. If you know that a team can be assigned at most one "site" group, then you can render that in a nicer way than if it could belong to multiple sites.

Right, which is why I **do** think it would be very useful to say that groups of type `site` must be a partition of all teams.

eldering: > > Ok, but in your examples I don't see groups that could overlap _and_ would naturally have the same type.
> 
> The first and third are exactly that. In the third example we would have groups for every country (with participants in the contest). Every team with a Swedish team member would be in the "Sweden" group, every team with a Canadian team member would be in the "Canada" group. So, if me and Tim make a team, our team would be in both those groups. Are you saying that these country groups would not be of the same type?

On the first, I don't really see why these need to have the same type, even have to have a type at all. For countries, I agree that's a possible use case. But then, will you also display multiple country flags on the scoreboard for that team?

> > I think this requirement would be useful, for example for a tool that would want to display the "site" of a team. If you know that a team can be assigned at most one "site" group, then you can render that in a nicer way than if it could belong to multiple sites.
> 
> Right, which is why I **do** think it would be very useful to say that groups of type `site` must be a partition of all teams.

I don't like it so much to attach special meaning to a particular group type `site`: that this one must be a partition of all teams. What about the following: next to `type` we introduce an attribute `label`. Then `type` should be used for groups that partition all teams, while `label` has no such restrictions.

niemela: > On the first, I don't really see why these need to have the same type, even have to have a type at all.

Sure. If we would specify that types must be proper partition we could just set these groups up as type-less.

> For countries, I agree that's a possible use case. But then, will you also display multiple country flags on the scoreboard for that team?

Yes? I'm not planning to implement it any time soon, this is all hypothetical, but yes, that's how I would do it. Also, it's not *completely* made up, we actually had a scoreboard like that some 15-20 years ago, for a contest with teams of 3 that were just freely formed. and not associated with universities or such.

> What about the following: next to `type` we introduce an attribute `label`. Then `type` should be used for groups that partition all teams, while `label` has no such restrictions.

I **really** don't like to have two types of types with different semantics. Also, the name label to me implies some display property rather than a grouping property.

niemela (dismissed):

Contest_API.md
eldering: I agree that `label` is maybe not good name and even useful enough to add. But I guess my point is: what is the value of `type` if it doesn't have some restrictions imposed on the groups that are part of a single type?
And I don't see a point in assigning a type to each group: if there's no clear reason to use `type`, I think it's perfectly fine to not assign a `type`, like with "has a graduate program" and "offers doctorates degrees".

Contest_API.md
```diff
@@ -773,12 +787,22 @@ JSON elements of group objects:
...
 
+#### Known group types
+
+The list below contains standardized identifiers for known group
+types. These identifiers should be used when the purpose
+of a group matches.
+
+| Type  | Description
+| ----- | ---------------------------------
+| site  | A physical location where teams are competing, e.g. the "Hawaii site". Teams generally should not be in more than one group of this type.
```
eldering, line 804: Remove "generally" as "should" already expresses the same. I would even prefer changing that into a "must" though.

niemela: I very much agree with you, but isn't this the very thing you argued against earlier?

> I don't like it so much to attach special meaning to a particular group type `site`: that this one must be a partition of all teams.

Maybe I misunderstood your argument there?

eldering: Well, I prefer making this a `must` over `should`, but I think it's even _better_ to make this a requirement encoded somehow in the specification of the `type` attribute, not on this specific type "site".

eldering (approved):

deboer-tim:
Replaced by PR #45 based on today's discussion.

PR #33, opened 2021-05-12
Add info about reading language configuration from the CAF.

nickygerritsen:
As suggested in #27.

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -267,6 +267,10 @@ The CCS must be able to read [groups.tsv](#groupstsv), an
 initialization file containing the list of groups to which the teams
 belong.
 
+The CCS must be able to read language configuration from the
+[Contest Archive Format](https://ccs-specs.icpc.io/contest_archive_format).
+See specifically the [language object](https://ccs-specs.icpc.io/contest_archive_format#language-object).
+
 ### Clock Synchronization
 
 The CCS must be able to synchronize its clock to an NTP server provided
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -267,8 +267,8 @@ The CCS must be able to read [groups.tsv](#groupstsv), an
 initialization file containing the list of groups to which the teams
 belong.
 
-The CCS must be able to read language configuration from the
-[Contest Archive Format](https://ccs-specs.icpc.io/contest_archive_format).
+The CCS must be able to read language configuration from an archive following
+the [Contest Archive Format](https://ccs-specs.icpc.io/contest_archive_format).
 See specifically the [language object](https://ccs-specs.icpc.io/contest_archive_format#language-object).
 
 ### Clock Synchronization


Contest_Control_System_Requirements.md
```diff
@@ -267,6 +267,10 @@ The CCS must be able to read [groups.tsv](#groupstsv), an
 initialization file containing the list of groups to which the teams
 belong.
 
+The CCS must be able to read language configuration from the
+[Contest Archive Format](https://ccs-specs.icpc.io/contest_archive_format).
```
niemela, line 271: Nitpick (on behalf of @clevengr also?). You can't read configuration **from** the Contest Archive **Format**. You can read configuration from an Archive, that can be saved in (or "using" or "following) the Contest Archive Format.

niemela (changes requested):

Contest_Control_System_Requirements.md
eldering: Another nitpick :-p What does it mean to read this from the contest archive? Can we then just ignore it?

nickygerritsen: Didn’t the old text for system.yaml had the same “issue”?

niemela: > Didn’t the old text for system.yaml had the same “issue”?

No? 

Firstly, was there even an old text directly referring to system.yaml?

Secondly, there is nothing wrong with saying "read data from system.yaml". It is a file, that follows a specified format. Saying "read data from CAF" would be the same as saying "read data from the system.yaml format", which makes no sense (and which I sure hope we never said).



> Another nitpick :-p What does it mean to read this from the contest archive? Can we then just ignore it after reading, and pipe it into `/dev/null`?

I think it would be better if we rewrote this thing to say that configuration is provided in this format (the CAF I guess), the CCS needs to be able to (manually and/or automatically) set up all the configurations allowed by that, and the CCS needs to be able to verify (or maybe some external system needs to be able to verify?) that the configuration has been done. That last thing could/should be done by exporting all the relevant data on the Contest API (or in a Contest Archive, which really is the exact same thing).

nickygerritsen: Feel free to push a suggested change to this branch. I agree we should be more clear

nickygerritsen: By the way are you suggesting we should add a requirement a CCS should be able to configure itself based on everything from the CAF, not only languages? Isn’t that a bigger change which might drop other requirements from this page?

niemela: Yes and yes.

nickygerritsen: So what is your suggestion? We only mention languages for now? Or do we already do everything? Or discuss on Monday?

niemela: Generally I don't think that a partial improvement should be blocked because there exists a larger more general improvement that we could also do. Taking iterative steps in the right direction is a great way to get somewhere, and requiring perfection for every step is a great way to get nowhere.

So, I'm absolutely fine only mentioning language for now. I also think we could discuss on Monday.

My nitpick about language use regarding the Contest Archive Format still stands though.

nickygerritsen: Yes I will address your nitpick before the meeting… I hope (since I agree with you, enjoy this moment 🤣)

eldering (approved):

niemela (approved):

PR #34, opened 2021-05-13
Add support for POST/PATCH/DELETE awards

deboer-tim:
Adds support for admin role to modify awards. This should be split into it's own feature (once we define how optional features should work) but for now I think it's ok to keep with awards (which are optional in the first place).

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1545,6 +1545,19 @@ For some common award cases the following IDs should be used.
 | group-winner-\<id>        | Current leader(s) in group \<id>. Empty if no team has scored.                                                             | Winner(s) of group \<id>          |                                                                         |
 | organization-winner-\<id> | Current leader(s) of organization \<id>. Empty if no team has scored.                                                      | Winner(s) of organization \<id>   | Not useful in contest with only one team per organization (e.g. the WF) |
 
+#### POST, PATCH, and DELETE awards
+
+Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
+
+The request must fail with a 4xx HTTP status code if any of the following happens:
+
+* The request doesn't include an award id.
+* A POST when the award id already exists.
+* A POST that is missing one of the required attributes (`citation` and `team_ids`).
+* A PATCH on an award id that doesn't exist.
+* A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
+* A DELETE on an award id that doesn't exist.
+
 #### Example
 
 Request:


nickygerritsen (approved):

niemela (approved):

Contest_API.md
```diff
@@ -1545,6 +1545,19 @@ For some common award cases the following IDs should be used.
 | group-winner-\<id>        | Current leader(s) in group \<id>. Empty if no team has scored.                                                             | Winner(s) of group \<id>          |                                                                         |
 | organization-winner-\<id> | Current leader(s) of organization \<id>. Empty if no team has scored.                                                      | Winner(s) of organization \<id>   | Not useful in contest with only one team per organization (e.g. the WF) |
 
+#### POST, PATCH, and DELETE awards
+
+Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
```
eldering, line 1550: I wasn't part of the discussion, but if I POST a new award, is the server supposed to update `team_ids` for it during the remainder of the contest? That would only be possible for award types the server knows about. And then how would the client know that the server knows about this award and it can thus expect updates?

nickygerritsen: It would not. If a client sends any awards, the server can stop generating their own awards.
The idea behind this PR is that a tool (like the award utility) can determine how awards are generated, at the end of a contest. Then it can “upload” them to a ccs or cds.
Before this change people often had to save a copy of the event feed to run a resolver, or “put it back in” the cds (if you wanted images). After this change you can run the resolver against a CCS including custom awards without going through strange hoops.

eldering: Maybe good to document this expected behavior then. So a (CCS) server might generate and update awards itself, but as soon as a client POSTs or PATCHes an award, the CCS won't update it anymore. Is that the current expected behavior?

nickygerritsen: I believe that is the idea, right @deboer-tim ?

niemela: > It would not. If a client sends any awards, the server can stop generating their own awards.

That's a *lot* stronger than anything we said. I definitely disagree.

Just to clarify, you are saying that if I use a tool to POST the "Best use of semicolon" award to the CCS can stop updating the `winner` award? 

First off, it **absolutely** can't just stop generating/updating, it would have to delete the award if it's not longer updating it (otherwise the `winner` would be listed as whoever was in the lead when we added the `semicolon` award. 

Secondly, adding novel awards (i.e. awards that the server has never heard of) should not change any behavior for awards the CCS is currently handling. I would find it very unexpected (and very bad) if adding the `semicolon` awards suddenly breaks all other award generation, if any.

Furthermore, for awards that the CCS **does** know about, I thing it should be allowed for the system to refuse the POST/PATCH/(DELETE?) (possibly depending on what access level is used). I.e. a CCS could decide that it will handle certain awards (winner and first-to-solve) comes to mind, and as such will not accept any updates to those prices. Maybe it should always accept DELETE though, and that should definitely also mean that it should stop generating the DELETEd award.

So maybe the semantics should be that the CCS has the responsibility for certain awards at the start of the contest (possibly an empty set) and if any of those get a POST/PATCH/DELETE the CCS can either refuse the update and keep the responsibility, or accept the update and drop the responsibility for future updates on that exact award.

In any case, we should probably add that to the spec?

eldering: > > It would not. If a client sends any awards, the server can stop generating their own awards.
> 
> That's a _lot_ stronger than anything we said. I definitely disagree.
> 
> Just to clarify, you are saying that if I use a tool to POST the "Best use of semicolon" award to the CCS can stop updating the `winner` award?

No, I meant (but maybe didn't say that so clearly) for any specific award. So in case you POST the "Best use of semicolon" award, the CCS would still update other awards as usual, but not the "Best use of semicolon" anymore going forward (if it already knew about it in the first place).

But indeed, I agree with your final statement: it should be clear what the responsibility is of the CCS and whether it can reject PATCHes or DELETEs to existing awards, or how it would need to handle these after accepting.

deboer-tim: I agree. I went to update the PR only to realize it was already merged. :) I tried to split this up as managed, overridable, and unmanaged awards, please see new PR #35.

PR #35, opened 2021-05-14
Managed awards

deboer-tim:
In PR #34 there was a conversation about whether a CCS allows awards to be overriden or not. Trying to address with this PR and three 'classes' of award ids: managed by the server and not changeable, assigned by default but overridable, and unmanaged/unknown.

There's still the question of whether a client can tell whether a particular award (potentially not even assigned yet) is managed by the server. I don't think this is critical though and maybe not an issue in practice, so I am not trying to address it here.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1549,6 +1549,10 @@ For some common award cases the following IDs should be used.
 
 Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
 
+Each award id may be managed by the server, overridable, or unmanaged (potentially even unknown). Managed awards are assigned by the server and any attempt to change them will fail. Overridable awards will be assigned by the server by default, but once it has been modified externally the server will stop assigning or updating them. Award ids that are unmanaged or unknown to the server are solely up to the client.
+
+For instance, a server may be configured to assign `winner` and `*-medal` awards as managed awards, and these cannot be overridden by clients. The same server may assign `first-to-solve-*` awards by default, but allow a client to override them. This server may have no support for assigning `group-winner-*` and these will only exist if a client adds them. The client can also add other or arbitrary awards like `first-submission-for-country`.
+
 The request must fail with a 4xx HTTP status code if any of the following happens:
 
 * The request doesn't include an award id.
@@ -1557,6 +1561,7 @@ The request must fail with a 4xx HTTP status code if any of the following happen
 * A PATCH on an award id that doesn't exist.
 * A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
 * A DELETE on an award id that doesn't exist.
+* A POST, PATCH, or DELETE on an award id that the server is managing.
 
 #### Example
 


nickygerritsen (approved):

Contest_API.md
```diff
@@ -1549,6 +1549,10 @@ For some common award cases the following IDs should be used.
 
 Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
 
+Each award id may be managed by the server, overridable, or unmanaged (potentially even unknown). Managed awards are assigned by the server and any attempt to change them will fail. Overridable awards will be assigned by the server by default, but once it has been modified externally the server will stop assigning or updating them. Award ids that are unmanaged or unknown to the server are solely up to the client.
```
niemela, line 1552: This taxonomy feels a bit unnatural. I.e. every award is exactly one of  `managed`, `overridable`, `unmanaged`, and (possibly) `unknown`.

What is the difference between unmanaged or unknown? It must be know to the server whether an award is `managed`, `overridable`, or `unmanaged`, so from the POV of the server it could never be `unknown`. Similarly, for the client there is initially no real way for the client to know the managed-state of any award. In any case, I don't see where it would ever make sense to classify some award as `unknown` vs (primarily) `unmanaged`.

Secondly, the name `overridable` when used in this context (with the other options being `managed` and `unmanaged`) clearly means "managed but that can be overridden". I.e. using plain language it would be reasonable to consider an `overridable` award to be (currently) managed by the server. I.e. it is "managed" but not `managed`. (I actually made this mistake and thought you had made an mistake in the last edit below, because I thought of an `overridable` award as "an award id that the server is managing".

I would prefer to simply group awards by whom they are managed by, either the server (i.e. `managed`) or the client (i.e. `unmanaged`). A server may or may not accept POST/PATCH/DELETE for awards it manages, **but** if it does then the client takes over managing those awards.

Also, note that there could be multiple clients writing to the server. So, and award could be managed not by the client doing the current request, but also not by the server, because it's managed by some other client.

In summary:
* The server must keep track of which awards it manages. 
* The server automatically manages all awards it creates. 
* Whenever the server accepts an POST/PATCH/DELETE it no longer manages that award, instead the client making the request does (except in the case of DELETE, where nobody manages the award since it no longer exists).
* The server must accept valid POST/PATCH/DELETEs on awards managed by the client making the request.
* The server may accept valid POST/PATCH/DELETEs on awards managed by it or other clients.
* (This implies that the server does not have to keep track of *which* client manages what, only whether an award is managed by the server itself or someone else).
* The client must assume that it manages an award after a successful POST/PATCH/DELETE.
* The client should/must assume that it no longer manages an award after a POST/PATCH/DELETE fails with 403 on that award.

Contest_API.md
```diff
@@ -1549,6 +1549,10 @@ For some common award cases the following IDs should be used.
...
+
+For instance, a server may be configured to assign `winner` and `*-medal` awards as managed awards, and these cannot be overridden by clients. The same server may assign `first-to-solve-*` awards by default, but allow a client to override them. This server may have no support for assigning `group-winner-*` and these will only exist if a client adds them. The client can also add other or arbitrary awards like `first-submission-for-country`.
```
niemela, line 1554: As written the first sentence seems to claim that it is a property of `winner` and `*-medal` that these could not be overridden by clients. But this is just an example, right? I would suggest changing "and these cannot be" to "and to not allow these to be". That said, this paragraph might change/disappear based on my earlier comment?

Contest_API.md
```diff
@@ -1557,6 +1561,7 @@ The request must fail with a 4xx HTTP status code if any of the following happen
 * A PATCH on an award id that doesn't exist.
 * A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
 * A DELETE on an award id that doesn't exist.
+* A POST, PATCH, or DELETE on an award id that the server is managing.
```
niemela, line 1564: As alluded to above "is managing" is unclear.

Based on my suggestion above this should be:
> The request **may** fail [...] 
> * A POST, PATCH, or DELETE on an award id that the client is not managing

...or something to that effect.

niemela (changes requested):

Contest_API.md
eldering: Sounds good to me!

A few comments:
* I would require that all server managed awards must exist. That is, it can't happen that a client POSTs a new award and then the server starts silently managing it. I guess this can be implicitly read from the text, but would be good to make explicit. Thus, all server managed awards exist from the start and if any PATCH or DELETE is sent on it, then it is not server managed anymore from that point on and never will be anymore, even if a client first sent a DELETE and then a POST to recreate it.
* Add that the server may accept POST requests on non-existing awards. (I guess in your summary it's not clear whether a non-existent award falls in the `unmanaged` category.)

niemela: >     I would require that all server managed awards must exist.

I disagree. An award could be managed by the server but not updated during the contest. The latter implies that it must not be present during the contest.

>     That is, it can't happen that a client POSTs a new award and then the server starts silently managing it.

If an award that does not exist (i.e. is not currently present) is actually managed by the server, **and** the server wants to retain that responsibility, it must reject the POST. If the POST is accepted the client must assume that it is now managing the award.

>     I guess this can be implicitly read from the text, but would be good to make explicit.

Actually, my last bullet point kinda implies the opposite. If, at some point, the client gets a 403 on a POST/PATCH/DELETE it should assume that it *no longer* manages that award. This implies that the server "started to silently manage it" again. I definitely think this should be allowed.

>     Thus, all server managed awards exist from the start

Absolutely not. This would imply that the server can never manage awards that are not updated during the contest.

>     [...] and if any PATCH or DELETE is sent on it, then it is not server managed anymore from that point on and never will be anymore, even if a client first sent a DELETE and then a POST to recreate it.

If any PATCH or DELETE is *accepted*, not simply sent.

>     Add that the server may accept POST requests on non-existing awards. (I guess in your summary it's not clear whether a non-existent award falls in the `unmanaged` category.)

Yes, that would be good to add. An non-existing award could be either managed by the server or not (a.k.a. `unmanaged`). Existence is not a requirement for being managed by the server. :)

eldering: > > I would require that all server managed awards must exist.
> 
> I disagree. An award could be managed by the server but not updated during the contest. The latter implies that it must not be present during the contest.

What would be the reason that it doesn't exist initially and then later does? I guess I don't understand this use case.

> > That is, it can't happen that a client POSTs a new award and then the server starts silently managing it.
> 
> If an award that does not exist (i.e. is not currently present) is actually managed by the server, **and** the server wants to retain that responsibility, it must reject the POST. If the POST is accepted the client must assume that it is now managing the award.

Agree.

> > I guess this can be implicitly read from the text, but would be good to make explicit.
> 
> Actually, my last bullet point kinda implies the opposite. If, at some point, the client gets a 403 on a POST/PATCH/DELETE it should assume that it _no longer_ manages that award. This implies that the server "started to silently manage it" again. I definitely think this should be allowed.

I guess that's possible, but I'd prefer to keep it simple and allow the client to assume that once an award is client-managed it will stay that way.

> > Thus, all server managed awards exist from the start
> 
> Absolutely not. This would imply that the server can never manage awards that are not updated during the contest.

I don't get this, but see above: I guess I don't understand a workflow you have in mind.

> > [...] and if any PATCH or DELETE is sent on it, then it is not server managed anymore from that point on and never will be anymore, even if a client first sent a DELETE and then a POST to recreate it.
> 
> If any PATCH or DELETE is _accepted_, not simply sent.

Agreed, I was not precise enough.

> > Add that the server may accept POST requests on non-existing awards. (I guess in your summary it's not clear whether a non-existent award falls in the `unmanaged` category.)
> 
> Yes, that would be good to add. An non-existing award could be either managed by the server or not (a.k.a. `unmanaged`). Existence is not a requirement for being managed by the server. :)

The concept of a managed, non-existing award is still unclear to me and seems to complicate things more than it might be useful.

nickygerritsen: > > > I would require that all server managed awards must exist.
> > 
> > 
> > I disagree. An award could be managed by the server but not updated during the contest. The latter implies that it must not be present during the contest.
> 
> What would be the reason that it doesn't exist initially and then later does? I guess I don't understand this use case.

This is a copy from the current awards endpoint:

> An award does not have to be present during the contest. However, if it is present, then it must be kept up to date during the contest. E.g. if "winner" will not be updated with the current leader during the contest, it must not be created until the award is awarded.

This makes a lot of sense if a CCS doesn't want to keep updating an award, but only generate it at the end of the contest (for example when setting `b` or when finalizing)

deboer-tim:
I've taken another pass at doing this based on the discussion on the last CLICS call. Since there is a lot of feedback here that doesn't apply, I've opened up new PR #43 and am closing this one.

PR #36, opened 2021-05-14
Fix merge error and general clean-up

niemela:
It seems that we accidentally duplicated the `compiler` and `runner` in some merge conflict. Fixed that and cleaned up the markup a bit.


niemela:
This is pure editing, I will merge without 2 approvals shortly.

PR #37, opened 2021-05-15
Move WF requirements from Contest API spec to CCS requirements docuement.

niemela:
This PR moves the information in all the "Source @ WF" columns and "Access restrictions at WF" sections.

Technically, the Contest API has never been required for a WF system, but now it is. The requirement for the Event Feed is removed from the CCS requirements document as it is now included via the Contest API by reference.

The `type` and `hidden` elements of the `/groups` endpoint are optional but were marked as "Source @ WF = CCS". It is not actually needed for the CCS to support them at WF, so I did not include them in the explicit list of optional endpoint and elements a WF system must support. Other than that, that list contains all endpoints and elements that are optional and had "Source @ WF = CCS".

The list of access restrictions have been edited very slightly to work in a common list (i.e. they can't refer to *this* endpoint), including some merging, but are otherwise kept as they were.

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -955,13 +955,34 @@ problem.
 
 ## Data Export
 
-### Event Feed
+### Contest API
 
-It is a requirement that the CCS provide an *external event feed*. This
-means that the CCS must have a mechanism for external processes to
-connect to the CCS and obtain dynamic real-time updates regarding the
-current state of the contest. The CCS event feed mechanism must comply
-with the [Event Feed specification](Event_Feed).
+It is a requirement that the CCS provide a compliant implmentation of
+the [Contest API](contest_api). In addition to all required endpoints
+and elements (i.e. the minimal compliant implmentation) the CCS must
+also provide the following endpoints and elements:
+
+| Endpoint         | Element                      |
+| :--------------- | :--------------------------- |
+| `/contests`      | `formal_name`                |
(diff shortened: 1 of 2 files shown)


Contest_Control_System_Requirements.md
```diff
@@ -955,13 +955,70 @@ problem.
...
+| `/organizations` | All required elements.       |
+| `/organizations` | `icpc_id`                    |
+| `/organizations` | `formal_name`                |
+| `/teams`         | `icpc_id`                    |
+| `/teams`         | `display_name`               |
+| `/teams`         | `organization_id`            |
+| `/teams`         | `group_ids`                  |
+| `/judgements`    | `max_run_time`               |
+| `/runs`          | `run_time`                   |
+| `/awards`        | All required elements.       |
```
nickygerritsen, line 985: Although the old API spuwde column clearly states the source was CCS, wasn’t this in practice the CDS, or more precisely the award util / resolver? If so, do we require this for a CCS at the WF? Although with the new POST etc changes we probably should

pehrsoderman (approved):

Contest_Control_System_Requirements.md
niemela: Yes, I was thinking exactly the same.

I suggest that we merge this PR as is, and then discuss if we want to change the status for awards at the next meeting. Does that sound good?

nickygerritsen: I agree

nickygerritsen (approved):

PR #39, opened 2021-05-17
Deprecate submissions.tsv

niemela:
The `submissions.tsv` output file can be completely replaced by the `/submissions` endpoint of the Contest API, or rather by the (submissions.json) file that you could download from that endpoint.

Also, it's not being referred to anywhere in the document, so even if it could not be replaced, it's not actually used.


nickygerritsen (approved):

nickygerritsen:
Lol, DOMjudge was not up to spec it seems, since we didn't do this afaik.

eldering (approved):

PR #42, opened 2021-05-20
SVG width/height clarification

deboer-tim:
Improve SVG width/height requirements for viewBox and avoid negative values otherwise. And Oxford comma!

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -265,11 +265,12 @@ equivalent JSON response snippets pointing to the same location:
   "href":"contests/wf14/submissions/187/files"
 ```
 
-For images, the supported mime types are image/png, image/jpeg and image/svg+xml.
+For images, the supported mime types are image/png, image/jpeg, and image/svg+xml.
 
 For images in SVG format, i.e. those having a mime type of image/svg+xml,
-the actual values of `width` and `height` don't matter, only the ratio
-between them.
+the values of `width` and `height` should be the viewBox width and height
+when possible, but otherwise the actual values don't matter as long as they
+are positive and represent the correct aspect ratio.
 
 If implementing support for uploading files pointed to by resource
 links, substitute the href element with a data element with a base64
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -268,7 +268,7 @@ equivalent JSON response snippets pointing to the same location:
 For images, the supported mime types are image/png, image/jpeg, and image/svg+xml.
 
 For images in SVG format, i.e. those having a mime type of image/svg+xml,
-the values of `width` and `height` should be the viewBox width and height
+the values of `width` and `height` should be the viewport width and height in pixels
 when possible, but otherwise the actual values don't matter as long as they
 are positive and represent the correct aspect ratio.
 


nickygerritsen (approved):

Contest_API.md
```diff
@@ -265,11 +265,12 @@ equivalent JSON response snippets pointing to the same location:
   "href":"contests/wf14/submissions/187/files"
 ```
 
-For images, the supported mime types are image/png, image/jpeg and image/svg+xml.
+For images, the supported mime types are image/png, image/jpeg, and image/svg+xml.
 
 For images in SVG format, i.e. those having a mime type of image/svg+xml,
-the actual values of `width` and `height` don't matter, only the ratio
-between them.
+the values of `width` and `height` should be the viewBox width and height
```
niemela, line 271: I have now spent more time then I'm willing to admit reading the SVG specs, and I *think* this is wrong. 

Also, when trying out my tentative understanding, the SVG viewing tools that I used displayed the images differently. The "glass half empty" view is that this proves that it is impossibly hard to get this right, so we have no chance. The "glass half full" view is that we are in good company, if established tools can't get the corner cases right, does it even matter if we do?

If a `width` and `height` are set on the SVG, that *is* the `width` and the `height` of the viewport, a.k.a. the image, regardless of the aspect ratio set in `viewBox`. The image will be stretched, padded or truncated, based on the value of `preserveAspectRatio` (`none` will stretch, `<align> meet` will pad and `<align> slice` will truncate). `xMidYMid meet` is the default, which produces an images that (kinda) seems to have the aspect ratio defined by `viewBox` since it just got some (invisible) padding added to it.

If there is no `width` and `height` set on the SVG then the aspect ratio is as defined by `viewBox`. Furthermore,  the `viewBox` values are in "user units", which I *think* can be thought of as "pixels" for our purposes.

If neither `width` and `height` nor `viewBox` then... I don't know. Bad server!

So, I *think* this is what would be correct:
- If `width` and `height` are set in pixels, use that.
- If `width` and `height` are set in some other unit(s) that can be translated to pixels, translate to pixels and use that.
- If `width` and `height` are set in some other unit(s) that can't be translated to pixels, fail I guess?
- If `width` and `height` are not set but `viewBox` is, and the width and height are integers, use that.
- If `width` and `height` are not set but `viewBox` is, and the width and height are not integers, multiply by some smallest power of 10(?) to make them integer and use that.
- If neither `width` and `height` nor `viewBox` are set, fail I guess?

Although, a simpler and mostly correct version would be to say that `width` and `height` should be the with and height of the SVG viewport in pixels.

niemela (changes requested):

Contest_API.md
deboer-tim: I totally agree with this, but I was trying to balance it against the complexity and the fact that every single SVG I've found that was built to be a document (i.e. not just an inline graphic) has had a viewBox set and no width/height/preserveAspectRatio... It seems like viewBox is typically what the SVG creator uses, and width/height is used by whoever wants to display it. And yes, there are edge cases where all of this fails, do we really want arbitrary data or to allow null width/height in file references?

But as you ended with, is the simplest-yet-still-correct fix just to change 'viewBox' to 'viewport'? My preference is definitely on the side of a simple "use what's there if you can find it" because I think that will cover >99.8% of cases, and I don't think any extra spec complexity is likely to make a significant difference on the rest.

niemela: > But as you ended with, is the simplest-yet-still-correct fix just to change 'viewBox' to 'viewport'?

Changing 'viewBox' to 'viewport' but also adding "in pixels". Yes, I think that has a very good correctness to complexity ratio.

> My preference is definitely on the side of a simple "use what's there if you can find it" because I think that will cover >99.8% of cases, and I don't think any extra spec complexity is likely to make a significant difference on the rest.

Agreed.

niemela (approved):

nickygerritsen (approved):

PR #43, opened 2021-05-20
Managed awards (2)

deboer-tim:
As discussed on the last CLICS call this is a clarification on managed awards and responsibilities. Hopefully this is simpler than PR #35 and clarifies the intent better.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1452,6 +1452,10 @@ For some common award cases the following IDs should be used.
 
 Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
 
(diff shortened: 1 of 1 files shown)


niemela (approved):

Contest_API.md
```diff
@@ -1460,6 +1464,7 @@ The request must fail with a 4xx HTTP status code if any of the following happen
 * A PATCH on an award id that doesn't exist.
 * A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
 * A DELETE on an award id that doesn't exist.
+* A POST, PATCH, or DELETE on an award id that the server is configured to manage exclusively.
```
eldering, line 1467: Maybe add `the server or another client is configured to manage exclusively.`

eldering (approved):

Contest_API.md
niemela: No, because it says *must* above.

PR #44, opened 2021-05-25
Consistent HTTP methods

deboer-tim:
As discussed on the mailing list, this PR adds some clarification to the HTTP methods section, returns the element for each update method, and makes the exist API (submissions, clarifications, awards) and examples match.

- POST may allow an id when in an admin role.
- POST always returns the Location header and initial state of element.
- PUT and PATCH return the current state of the element.
- POST submissions and clarifications updated to match the above.
- /awards adds PUT, made consistent with above, id not required, and URLs fixed.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -116,8 +116,12 @@ their expected behavior, if implemented.
     It can be used to request a whole collection or a specific element.
   - `POST`
     Create a new element. This can only be called on a collection
-    endpoint. No `id` attribute should be specified as it is up to the
-    server to assign one, which is returned in the location header.
+    endpoint. When using most [roles](#Roles) the `id` attribute may not
+    be specified as it is up to the server to assign one. When using an
+    `admin` role you may specify a new, unused id.
+    The response will contain a `Location` header pointing to the newly
+    created element and the response body will contain the initial
+    state of the element.
   - `PUT`
     Replaces a specific element. This method is idempotent and can only
     be called on a specific element and replaces its contents with the
@@ -125,18 +129,23 @@ their expected behavior, if implemented.
     updates are allowed. The `id` attribute cannot be changed: it does
     not need to be specified (other than in the URL) and if specified
     different from in the URL, a `409 Conflict` HTTP code should be
-    returned.
+    returned. The response body will contain the updated state of the
+    element.
   - `PATCH`
     Updates/modifies a specific element. Similar to `PUT` but allows
     partial updates by providing only that data, for example:
     `PATCH  https://example.com/api/contests/wf14/teams/10`
     with JSON contents
-    `{"name":"Our cool new team name"}`
-    No updates of the `id` attribute are allowed either.
+    `{"name":"Our cool new team name"}`.
+    The `id` attribute cannot be changed: it does
+    not need to be specified (other than in the URL) and if specified
+    different from in the URL, a `409 Conflict` HTTP code should be
+    returned. The response body will contain the updated state of the
+    element.
   - `DELETE`
     Delete a specific element. Idempotent, but may return a 404 status
     code when repeated. Any provided data is ignored. Example:
-    `DELETE  https://example.com/api/contests/wf14/teams/8`
+    `DELETE  https://example.com/api/contests/wf14/teams/8`.
     Note that deletes must keep [referential
     integrity](#referential-integrity) intact.
 
@@ -1099,7 +1108,8 @@ The request must fail with a 400 error code if any of the following happens:
   files, it is too big, etc.
 * The provided `id` already exists or is otherwise not acceptable.
 
-The response will be the ID of the newly added submission.
+The response will contain a `Location` header pointing to the newly created submission
+and the response body will contain the initial state of the submission.
 
 Performing a `POST` by any other roles than `admin` and `team` is not supported.
 
@@ -1182,19 +1192,28 @@ Request data:
 
 ```json
 {
-   "language_id": "1-java",
-   "problem_id": "10-asteroids",
-   "team_id": "123",
-   "time": "2014-06-25T11:22:05.034+01",
-   "entry_point": "Main",
-   "files": [{"data": "<base64 string>"}]
+   "language_id":"1-java",
+   "problem_id":"10-asteroids",
+   "team_id":"123",
+   "time":"2014-06-25T11:22:05.034+01",
+   "entry_point":"Main",
+   "files":[{"data": "<base64 string>"}]
 }
 ```
 
 Returned data:
 
 ```json
-"187"
+{
+   "id":"187",
+   "language_id":"1-java",
+   "problem_id":"10-asteroids",
+   "team_id":"123",
+   "time":"2014-06-25T11:22:05.034+01",
+   "contest_time":"1:22:05.034",
+   "entry_point":"Main",
+   "files":[{"href":"contests/wf14/submissions/187/files","mime":"application/zip"}]
+}
 ```
 
 ### Judgements
@@ -1335,7 +1354,8 @@ The request must fail with a 4xx HTTP status code if any of the following happen
   visible to the role that's submitting.
 * The provided `id` already exists or is otherwise not acceptable.
 
-The response will be the ID of the newly added clarification.
+The response will contain a `Location` header pointing to the newly created clarification
+and the response body will contain the initial state of the clarification.
 
 Performing a `POST` by any roles other than `admin` and `team` is not supported.
 
@@ -1387,16 +1407,23 @@ Request data:
 
 ```json
 {
-   "problem_id": "10-asteroids",
-   "from_team_id": "34",
-   "text": "Can I assume the asteroids are round?"
+   "problem_id":"10-asteroids",
+   "from_team_id":"34",
+   "text":"Can I assume the asteroids are round?"
 }
 ```
 
 Returned data:
 
 ```json
-"187"
+{
+   "id":"clar-43",
+   "problem_id":"10-asteroids",
+   "from_team_id":"34",
+   "text":"Can I assume the asteroids are round?",
+   "time":"2017-06-25T11:59:47.543+01",
+   "contest_time":"1:59:47.543"
+}
 ```
 ### Awards
 
@@ -1448,9 +1475,9 @@ For some common award cases the following IDs should be used.
 | group-winner-\<id>        | Current leader(s) in group \<id>. Empty if no team has scored.                                                             | Winner(s) of group \<id>.                 |
 | organization-winner-\<id> | Current leader(s) of organization \<id>. Empty if no team has scored.                                                      | Winner(s) of organization \<id>.          | Not useful in contest with only one team per organization (e.g. the WF).
 
-#### POST, PATCH, and DELETE awards
+#### POST, PUT, PATCH, and DELETE awards
 
-Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
+Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, replace one with PUT, PATCH one or more attributes, or DELETE an existing award. POSTs should include the `id` when creating new awards of the known award types above; if the id is omitted the server will assign one.
 
 The server may be configured to manage (assign or update) some award ids, and may block clients from modifying them. However, if a client is able to modify an award it must assume that it is responsible for managing that award id unless and until it sees an indication that something else is now managing that award - either a change that it did not request, or a future modification fails.
 
@@ -1458,13 +1485,12 @@ For example, the server may be configured to assign the `winner` award and not a
 
 The request must fail with a 4xx HTTP status code if any of the following happens:
 
-* The request doesn't include an award id.
-* A POST when the award id already exists.
-* A POST that is missing one of the required attributes (`citation` and `team_ids`).
-* A PATCH on an award id that doesn't exist.
+* A POST that includes an id that already exists.
+* A PUT, PATCH, or DELETE on an award that doesn't exist.
+* A POST or PUT that is missing one of the required attributes (`citation` and `team_ids`).
 * A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
-* A DELETE on an award id that doesn't exist.
-* A POST, PATCH, or DELETE on an award id that the server is configured to manage exclusively.
+* A PUT, PATCH, or DELETE that includes an award id that don't match the id in the url.
+* A POST, PUT, PATCH, or DELETE on an award id that the server is configured to manage exclusively.
 
 #### Example
 
@@ -1493,23 +1519,27 @@ Request data:
 
 Request:
 
-` PATCH https://example.com/api/contests/wf14/awards`
+` PUT https://example.com/api/contests/wf14/awards/best-costume`
 
 Request data:
 
 ```json
-{"id":"best-costume","citation":"Best team costumes"}
+{"citation":"Best team costuems","team_ids":["24"]}
 ```
 
 Request:
 
-` DELETE https://example.com/api/contests/wf14/awards`
+` PATCH https://example.com/api/contests/wf14/awards/best-costume`
 
 Request data:
 
 ```json
-{"id":"best-costume"}
+{"citation":"Best team costumes"}
 ```
+
+Request:
+
+` DELETE https://example.com/api/contests/wf14/awards/best-costume`
 
 ### Commentary
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -137,17 +137,18 @@ their expected behavior, if implemented.
 
   - `DELETE`
     Delete a specific element. Idempotent, but will return a `404` error
-    code when repeated. Any provided data is ignored and there is no response body.
+    code when repeated. Any provided data is ignored, and there is no response body if successful.
     Example: `DELETE https://example.com/api/contests/wf14/teams/8`.
     Note that deletes must keep [referential integrity](#referential-integrity) intact.
 
 #### Success, Failure, and HTTP Responses
 
 Standard [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
-returned to indicate success or failure. If successful the response body of every method
-(except for DELETE, and GET on a collection) will contain the current (updated) state of the element.
+returned to indicate success or failure. If successful DELETE will have no response body,
+GET on a collection will return the collection, and every other method will contain the
+current (updated) state of the element.
 
-If a POST, PUT, PATCH would cause any of the following issues it will fail, in addition to any endpoint or element-specific requirements:
+If a POST, PUT, PATCH would cause any of the following issues it must fail, in addition to any endpoint or element-specific requirements:
 
 * A PATCH on an `id` that doesn't exist. Will return a 404 error code.
 * A PUT or PATCH containing an id that does not match the URL. Will return a 409 error code.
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

Contest_API.md
```diff
@@ -1448,23 +1475,22 @@ For some common award cases the following IDs should be used.
...
 | organization-winner-\<id> | Current leader(s) of organization \<id>. Empty if no team has scored.                                                      | Winner(s) of organization \<id>.          | Not useful in contest with only one team per organization (e.g. the WF).
 
-#### POST, PATCH, and DELETE awards
+#### POST, PUT, PATCH, and DELETE awards
 
-Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, PATCH one or more attributes, or DELETE an existing award. All requests must include the id attribute.
+Clients with the `admin` role may make changes to awards using the normal [HTTP methods](#http-methods) as specified above. Specifically, they can POST new awards, replace one with PUT, PATCH one or more attributes, or DELETE an existing award. POSTs should include the `id` when creating new awards of the known award types above; if the id is omitted the server will assign one.
```
eldering, line 1480: Does it make sense for the server to generate an `id` if none is given? I'd find it more natural that the server is allowed to reject a POST without an `id`. The same might also apply to POSTs to problems and languages (although we don't support that yet).

Contest_API.md
```diff
@@ -1448,23 +1475,22 @@ For some common award cases the following IDs should be used.
...
-* A POST when the award id already exists.
-* A POST that is missing one of the required attributes (`citation` and `team_ids`).
-* A PATCH on an award id that doesn't exist.
+* A POST that includes an id that already exists.
+* A PUT, PATCH, or DELETE on an award that doesn't exist.
+* A POST or PUT that is missing one of the required attributes (`citation` and `team_ids`).
 * A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
-* A DELETE on an award id that doesn't exist.
-* A POST, PATCH, or DELETE on an award id that the server is configured to manage exclusively.
+* A PUT, PATCH, or DELETE that includes an award id that don't match the id in the url.
```
eldering, line 1492: Misspelling: `does not`.

Contest_API.md
```diff
@@ -1493,24 +1519,28 @@ Request data:
...
 Request:
 
-` PATCH https://example.com/api/contests/wf14/awards`
+` PUT https://example.com/api/contests/wf14/awards/best-costume`
 
 Request data:
 
 ```json
-{"id":"best-costume","citation":"Best team costumes"}
+{"citation":"Best team costuems","team_ids":["24"]}
```
eldering, line 1527: Nitpick: should probably be `costumes`.

Contest_API.md

Contest_API.md
deboer-tim: I was trying to make it look like they misspelled costumes and then fixed it, but wasn't obvious enough. New commit changes it to costumes and then renames citation to cosplay, hopefully this is better.

Contest_API.md
deboer-tim: I was assuming the id of a random/unknown award didn't matter, so a generated one was just as good as the client picking one. Are you totally against generated award ids, or just best practice to include one?

Contest_API.md
```diff
@@ -116,27 +116,36 @@ their expected behavior, if implemented.
     It can be used to request a whole collection or a specific element.
   - `POST`
     Create a new element. This can only be called on a collection
-    endpoint. No `id` attribute should be specified as it is up to the
-    server to assign one, which is returned in the location header.
+    endpoint. When using most [roles](#Roles) the `id` attribute may not
+    be specified as it is up to the server to assign one. When using an
+    `admin` role you may specify a new, unused id.
```
niemela, line 121: 1. If you specify an `id` in your `POST`, you could have used a `PUT` instead, and as a general rule one should prefer `PUT` over `POST`, since `PUT` is idempotent, and `POST` is not. (This general rule is definitely being argued on the Internet though...)
One could make the case that the difference between `POST`ing with an `id` and `PUT`ing to the same `id` is that the former is a suggestion and the latter a requirement. `PUT`ing to an `id` should only ever either create or update that `id` or be rejected. `POST`ing *could* (and maybe should, I'm not sure?) create a different `id`.

2. Why are we limiting to "a new, unused id". In fact, using `POST` to update an existing `id` would make as much sense.

3. It's not true that you always may specify an `id` (unused or otherwise) when you are `admin`, it depends on the endpoint.

In summary:

I think that I think that whenever you are supplying an `id` with a `POST` it would be better to instead use the equivalent `PUT`. But that doesn't mean that we couldn't *also* allow the `POST`.

I think we should say less here. Maybe something like:
> Typically the `id` attribute must not be specified as it is up to the server to assign one.

Contest_API.md
```diff
@@ -116,27 +116,36 @@ their expected behavior, if implemented.
...
+    The response will contain a `Location` header pointing to the newly
+    created element and the response body will contain the initial
+    state of the element.
   - `PUT`
     Replaces a specific element. This method is idempotent and can only
```
niemela, line 126: A `PUT` can absolutely create an element. As noted above, I would argue that this would be the preferred way to create an element with a known `id`.

Contest_API.md
```diff
@@ -116,27 +116,36 @@ their expected behavior, if implemented.
...
     be called on a specific element and replaces its contents with the
     data provided. The payload data must be complete, i.e. no partial
     updates are allowed. The `id` attribute cannot be changed: it does
     not need to be specified (other than in the URL) and if specified
     different from in the URL, a `409 Conflict` HTTP code should be
```
niemela, line 131: Why do we want to allow not specifying the `id`? Is there any use case where this would be beneficial? If not. I would prefer to require it. (That is kinda what the HTTP specs says after all)

Contest_API.md
```diff
@@ -116,27 +116,36 @@ their expected behavior, if implemented.
...
   - `PATCH`
     Updates/modifies a specific element. Similar to `PUT` but allows
     partial updates by providing only that data, for example:
     `PATCH  https://example.com/api/contests/wf14/teams/10`
     with JSON contents
-    `{"name":"Our cool new team name"}`
-    No updates of the `id` attribute are allowed either.
+    `{"name":"Our cool new team name"}`.
+    The `id` attribute cannot be changed: it does
+    not need to be specified (other than in the URL) and if specified
```
niemela, line 141: Same as above: I think `id`s should be required.

Contest_API.md
niemela: Even more natural would be to `PUT` if you must specify the `id`.

Contest_API.md
```diff
@@ -1448,23 +1475,22 @@ For some common award cases the following IDs should be used.
...
 
 For example, the server may be configured to assign the `winner` award and not allow any client to modify it. The same server may assign `*-medal` awards by default, but allow clients to modify them. Once a client modifies any of the `*-medal` awards, it is responsible for updating it if anything changes. Likewise, the client could add any arbitrary awards like `first-submission-for-country-*` and would be responsible for managing these.
 
 The request must fail with a 4xx HTTP status code if any of the following happens:
 
-* The request doesn't include an award id.
-* A POST when the award id already exists.
-* A POST that is missing one of the required attributes (`citation` and `team_ids`).
-* A PATCH on an award id that doesn't exist.
+* A POST that includes an id that already exists.
```
niemela, line 1488: Not sure if that should be a "must fail"?

niemela (changes requested):

Contest_API.md
nickygerritsen: 1. So you are also suggesting to change POST to `submissions` to be a PUT to `submissions/<id>` if you supply an ID? This does seem to be more in line with the RFC, but I am guessing that if we go this way that DOMjudge will just accept post POST and PUT to both the collection as well as a specific ID (and only verify if the ID matches anything you supply in the body
2. I am not a fan at all of updating the ID of an existing entity. How would that go in the event feed? This also means you need to propagate the ID to any entities that uses it (in the event feed). Please let's not allow this. Also, should this then not be a `PUT`?
3. I agree with your third point

Contest_API.md
nickygerritsen: As said above: I'm fine with requiring the ID but I am not fine with allowing it to be different than what it was before.

Contest_API.md
nickygerritsen: I think for `PATCH` this is not true.

Contest_API.md
deboer-tim: 1/2. When we originally wrote this section I think the intent was clearly that POST was for creating an element and PUT was for replacing an existing element. I'm ok changing that and your points about the HTTP spec in 1 and 2 are correct, but you're also pointing how many ways POST and PUT are the same and I don't really want to make these methods basically equivalent or interchangeable either.

Reading the HTTP spec, the difference I get is POST is 'could you please accept this resource somewhere' and PUT is 'I want this resource here'. What if we left POST for cases where you don't have or know the id and it must not be included (e.g. teams submitting), and PUT for cases where you know the id and just want to set or replace the object (e.g. admins or some of the proxy cases)? To me this would provide some good differentiation in purpose and clarity to the behaviour.

If we went that way it would make total sense to not allow id in POST, require it in PUT (since you are replacing a 'whole' object), and have it optional in PATCH (since you only need to include any attributes you want to change).

As Nicky said I think PUT and PATCH should fail if the id in the url and attribute don't match, as that is changing the URL of the resource. It's simpler to just require a DELETE and then a PUT.

3. Agreed, I worded this badly.

Contest_API.md
deboer-tim: I don't know why this was done originally, but I agree it should be required in PUT, and I agree with Nicky that it must match the URI.

Contest_API.md
deboer-tim: Agreed with Nicky, PATCH should only be required to include attributes you want to change.

Contest_API.md
niemela: >     1. DOMjudge will just accept post POST and PUT to both the collection as well as a specific ID

Accepting a `PUT` to the collection is absolutely wrong (unless you're `PUT`ing the actual entire collection). Accepting a `POST` to the entity is arguably wrong. So, please don't do that.

>     2. I am not a fan at all of updating the ID of an existing entity.

Sorry, I was being very unclear. You should *not* be able to update the `id` of en existing object, what I meant by "update an existing `id`" was to update the object identified by some `id`.

> If we went that way it would make total sense to not allow id in POST, require it in PUT (since you are replacing a 'whole' object), and have it optional in PATCH (since you only need to include any attributes you want to change).

That sounds good to me.

> As Nicky said I think PUT and PATCH should fail if the id in the url and attribute don't match, as that is changing the URL of the resource. It's simpler to just require a DELETE and then a PUT.

Definitely agreed.

Contest_API.md
niemela: Yeah, you're absolutely right.

Contest_API.md
nickygerritsen: > > ```
> > 1. DOMjudge will just accept post POST and PUT to both the collection as well as a specific ID
> > ```
> 
> 
> Accepting a `PUT` to the collection is absolutely wrong (unless you're `PUT`ing the actual entire collection). Accepting a `POST` to the entity is arguably wrong. So, please don't do that
PUTting to the collection is also wrong. So if we both promise to never to it we can’t verify we do 😅. We might add a check
> 
> 
> > ```
> > 2. I am not a fan at all of updating the ID of an existing entity.
> > ```
> 
> 
> Sorry, I was being very unclear. You should _not_ be able to update the `id` of en existing object, what I meant by "update an existing `id`" was to update the object identified by some `id`.
👍
> 
> 
> > If we went that way it would make total sense to not allow id in POST, require it in PUT (since you are replacing a 'whole' object), and have it optional in PATCH (since you only need to include any attributes you want to change).
> 
> 
> That sounds good to me.
> 
> 
> > As Nicky said I think PUT and PATCH should fail if the id in the url and attribute don't match, as that is changing the URL of the resource. It's simpler to just require a DELETE and then a PUT.
> 
> 
> Definitely agreed.

niemela: > PUTting to the collection is also wrong. So if we both promise to never to it we can’t verify we do 😅. We might add a check

Ah, I see what you're saying. Yes, that makes sense. :)

Contest_API.md
```diff
@@ -114,36 +114,58 @@ their expected behavior, if implemented.
...
   - `DELETE`
-    Delete a specific element. Idempotent, but may return a 404 status
-    code when repeated. Any provided data is ignored. Example:
-    `DELETE  https://example.com/api/contests/wf14/teams/8`
-    Note that deletes must keep [referential
-    integrity](#referential-integrity) intact.
+    Delete a specific element. Idempotent, but will return a `404` error
+    code when repeated. Any provided data is ignored and there is no response body.
+    Example: `DELETE https://example.com/api/contests/wf14/teams/8`.
+    Note that deletes must keep [referential integrity](#referential-integrity) intact.
```
niemela, line 142: It's (a tiny bit) unclear if this note defines a requirement on the client or the server. The answer is, "the server", right? If so, this implies that the server must implement cascade or deny semantics, but it's free to choose which.

Contest_API.md
```diff
@@ -114,36 +114,58 @@ their expected behavior, if implemented.
...
+
+#### Success, Failure, and HTTP Responses
+
+Standard [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
+returned to indicate success or failure. If successful the response body of every method
+(except for DELETE, and GET on a collection) will contain the current (updated) state of the element.
```
niemela, line 148: How is the `GET` an exception?

`GET` on a collection will return the current state of the element(s).

Contest_API.md
```diff
@@ -114,36 +114,58 @@ their expected behavior, if implemented.
...
+
+If a POST, PUT, PATCH would cause any of the following issues it will fail, in addition to any endpoint or element-specific requirements:
```
niemela, line 150: I think this should be "must" instead of "will"? Same below.

Contest_API.md
```diff
@@ -114,36 +114,58 @@ their expected behavior, if implemented.
...
+
+* A PATCH on an `id` that doesn't exist. Will return a 404 error code.
+* A PUT or PATCH containing an id that does not match the URL. Will return a 409 error code.
+* A missing required attribute.
+* An attribute that must not be provided is provided.
+* An attribute type that is incorrect or otherwise invalid (e.g. non-nullable attribute set to null).
+* A reference to another element is invalid (to maintain [referential integrity](#referential-integrity)).
```
niemela, line 157: Can it be invalid in other ways than "missing"? If not, should we say "missing" instead?

Contest_API.md
```diff
@@ -114,36 +114,58 @@ their expected behavior, if implemented.
...
+
+A DELETE will fail in the following conditions, in addition to any endpoint or element-specific requirements:
+
+* Deleting an element `id` that doesn't exist. Will return a 404 error code.
+* If the element being deleted is referenced by another element (to maintain [referential integrity](#referential-integrity)).
```
niemela, line 162: I think it should be allowed for the server to implement cascade instead. I.e., it should not be a requirement for this `DELETE` to fail.

Contest_API.md
```diff
@@ -427,19 +449,17 @@ Countdown is resumed by setting a new `start_time` and resetting
 
 #### PATCH start\_time
 
-To replace the *Contest Start Interface*, at the ICPC World
-Finals, an API provided by a CCS or CDS implementing this specification
-must have a role that has the ability to clear or set the contest start
-time via a PATCH method.
+Implementations must have an admin role that has the ability to clear or set the
+contest start time via a PATCH method.
```
niemela, line 453: I prefer "a role that..." over "an admin role that..."

Contest_API.md
```diff
@@ -1448,23 +1488,22 @@ For some common award cases the following IDs should be used.
...
 
-The request must fail with a 4xx HTTP status code if any of the following happens:
+The request must fail with a 4xx error code if any of the following happens:
 
-* The request doesn't include an award id.
-* A POST when the award id already exists.
-* A POST that is missing one of the required attributes (`citation` and `team_ids`).
-* A PATCH on an award id that doesn't exist.
+* A POST that includes an id.
+* A PUT, PATCH, or DELETE on an award that doesn't exist.
```
niemela, line 1502: PUT on an award that doesn't exist *is* allowed.

Contest_API.md
```diff
@@ -1448,23 +1488,22 @@ For some common award cases the following IDs should be used.
...
+* A POST or PUT that is missing one of the required attributes (`citation` and `team_ids`).
 * A PATCH that contains an invalid attribute (e.g. null `citation` or `team_ids`).
-* A DELETE on an award id that doesn't exist.
-* A POST, PATCH, or DELETE on an award id that the server is configured to manage exclusively.
+* A PUT, PATCH, or DELETE that includes an award id that doesn't match the id in the url.
```
niemela, line 1505: That can't happen for a `DELETE`.

niemela (changes requested):
I think this is a big improvement. My comments are all fairly minor.

I still don't love how we mix in requirements on exactly how the different roles must work, in the body of the spec. But, we can discuss more about that later...

Contest_API.md
```diff
@@ -114,36 +114,58 @@ their expected behavior, if implemented.
...
+    with JSON contents `{"name":"Our cool new team name"}`.
+
   - `DELETE`
-    Delete a specific element. Idempotent, but may return a 404 status
-    code when repeated. Any provided data is ignored. Example:
-    `DELETE  https://example.com/api/contests/wf14/teams/8`
-    Note that deletes must keep [referential
-    integrity](#referential-integrity) intact.
+    Delete a specific element. Idempotent, but will return a `404` error
+    code when repeated. Any provided data is ignored and there is no response body.
```
eldering, line 140: I think the body should contain a message in case of an error, right?

Contest_API.md
eldering: Maybe in case of a clarification `reply_to_id` that points to another clarification that the senders of this clarification don't have access to. I'd be happy to leave as is, or just say "typically when the referenced element is missing".

Contest_API.md
eldering: This is a WF specific requirement, so I think we should leave this optional and then require it in the CCS requirements document.

Contest_API.md
```diff
@@ -1061,34 +1081,37 @@ zip archive. These must be stored directly from the root of the zip
...
-  must match the ID of the team associated with the request. `time` will then always
-  use the current time as determined by the server. The CCS will determine an `id`.
-* If an `id` is supplied, the client should make sure it is unique, i.e. not used
-  yet on the CCS. The client should normally not supply `id`, but let it be determined
-  by the server. However, for example in a setup with a central CCS with satellite sites
+* If the CCS supports a `team` role they will only have access to `POST`, and `time`
+  must not be provided when using this role. `time` will always be
+  set to the current time as determined by the server. `team_id` may be provided but then
+  must match the ID of the team associated with the request.
+* Simple proxys using an `admin` role should also use `POST` to let the server
```
eldering, line 1103: Nitpick typo: `proxies`

Contest_API.md
nickygerritsen: Hmmmm but the pausing of contest time is generally useful (and has been used outside of WF for sure). I’m not sure what part is WF specific?

eldering: Well, it's useful, but it's not critical for functioning. So I don't think this should be a requirement in general.

nickygerritsen: I think this is one that is clearly useful as an optional feature which we can do when we have the role/feature idea?

Contest_API.md
deboer-tim: Agreed. New commit changes it to a 'must implement fail or cascading delete', see if that works.

Contest_API.md
deboer-tim: I was trying to say there are two places that won't return one element. Reworded in new commit, see if that's better.

Contest_API.md

Contest_API.md

Contest_API.md
deboer-tim: New commit removes the 'admin'.

This was required before and I wasn't trying to change that via this PR. I totally agree we should implement optional features and this should be one of them (and required for finals).

Contest_API.md

Contest_API.md

Contest_API.md
deboer-tim: Reworded in new commit to add 'if successful'.

Contest_API.md

Contest_API.md
niemela: > pausing of contest time is generally useful

I would argue that pausing of contest time is a general anti-feature... but that's another discussion :).

> This was required before and I wasn't trying to change that via this PR. I totally agree we should implement optional features and this should be one of them (and required for finals).

I agree that this is outside the score of this PR. I also agree with @eldering and @deboer-tim that it should be moved to teh CCS requirements document.

niemela (approved):

nickygerritsen (approved):

PR #45, opened 2021-05-31
Squashed and merged groups - site type & scoreboards

deboer-tim:
I tried other approaches, but only clean one I could figure out was to just resubmit PR #32 as a new PR. Removed the word 'strongly', everything else should be identical and based on the current main branch.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -836,7 +836,7 @@ JSON elements of group objects:
 | icpc\_id | string  | no        | yes       | External identifier from ICPC CMS.
 | name     | string  | yes       | no        | Name of the group.
 | type     | string  | no        | yes       | Type of this group.
-| hidden   | boolean | no        | yes       | | if the group is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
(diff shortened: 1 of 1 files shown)


eldering (approved):

Contest_API.md
```diff
@@ -822,7 +836,17 @@ JSON elements of group objects:
 | icpc\_id | string  | no        | yes       | External identifier from ICPC CMS.
 | name     | string  | yes       | no        | Name of the group.
 | type     | string  | no        | yes       | Type of this group.
-| hidden   | boolean | no        | yes       | If group should be hidden from scoreboard. Defaults to false if missing.
+| hidden   | boolean | no        | yes       | | if the group is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
```
niemela, line 839: There is an extra pipe here.
```suggestion
| hidden   | boolean | no        | yes       | if the group is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
```

niemela (changes requested):

niemela (approved):

PR #46, opened 2021-06-02
Move hidden property from group to team

niemela:
Move the `hidden` property from the `groups` endpoint to the `teams` endpoint.


nickygerritsen (approved):

deboer-tim (approved):
Capitalize "if"? (I missed that in my last PR, didn't realize you had changed them all)

Approving, but would like a quick chat about backwards-compatibility on the next call.

PR #52, opened 2021-10-01
Remove "removing time intervals".

meisterT:
IMO, we should update the spec and reflect reality here. Not all CCSs implement removed time intervals. It is a basically untested feature which is even more dangerous than not having it.

In cases where it really would have been applied (think about fire alarm, bomb threat etc) the contest is impacted in such a way anyway that **if** you want to do something to make it more fair, you can use the "adding penalty time" functionality instead.

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -376,36 +376,6 @@ time.
 
 ### Adjusting for Exceptional Circumstances
 
-#### Removing time intervals
-
-It must be possible to, potentially retroactively, specify time
-intervals that will be disregarded for the purpose of scoring. The time
-during all such intervals will not be counted towards a team's penalty
-time for solved problems. Beginning and end of time intervals are given
-in wall-clock time.
-
-Note that removing a time interval changes the wall-clock time when the
-contest ends, as the duration of the contest in
-[contest.yaml](#contestyaml) is specified in contest time.
-
-Removing the interval between time T<sub>0</sub> and T<sub>1</sub>,
-where T<sub>0</sub> ≤ T<sub>1</sub>, means that all submissions received
-between T<sub>0</sub> and T<sub>1</sub> will have the same contest time
-and that the time for all submissions received after T<sub>1</sub> will
-have a contest time that is T<sub>1</sub>-T<sub>0</sub> lower than if
-the interval was not removed.
-
-The removal of a time interval must be reversible. This means that if a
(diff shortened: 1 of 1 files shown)


meisterT:
cc @niemela @eldering @clevengr

nickygerritsen (approved):
I totally agree with this.

pehrsoderman (approved):

eldering:
I completely support this!

Just a minor correction: the mention of removed intervals at https://github.com/icpc/ccs-specs/pull/52/files#L889 should also be removed.

eldering (approved):

nickygerritsen (approved):

niemela:
There has been no progress on this discussion for the past 3 years. Has the arguments for or against evolved any? If not, I would suggest we close this as wontfix. If things change it's easy enough to open it up again.

As one potential argument, I can point out that this feature has still never been used, and for the past several years, I don't even think we have tested it seriously (as we did do for the first several years). Do we think that the current CCS:s implementation of this still work? @clevengr @meisterT (and others)

meisterT:
Instead of keeping it open, can we turn this around and say: please speak up if you want to keep it? (Maybe @clevengr but it's not clear to me)

As a sanity check, we can and maybe should ask the judges what their point of view is. 

If we actually want to keep it we have to make clear how to communicate the removal of a time interval to shadow and other downstream clients.

nickygerritsen:
While shadowing the practice session of GNY2024 a few weeks ago PC^2 "stopped" their clock for 69 seconds without telling DOMjudge.

Their argument was "you can see it in the feed since contest times and absolute times started to drift", but since DOMjudge only stores one of these times, I had to write a small python script to detect where the 69 seconds were "lost" (note they didn't know how long the pause was, I had to determine this) and then insert a removed interval.

The good news is that that worked. The (IMHO) bad news is that indeed there is no proper spec way to tell us, especially when it happens in the future.

I agree with Tobi that we should at least discuss with the judges and if this has to stay decide on how to do it properly

deboer-tim:
Agreed with @niemela that we need to just make a call and close this either way. API support for doing this could be as simple as a pause object with start and optional end - either that's part of a consistent spec, or we remove the reference and agree on the API elsewhere for those that choose to support it as an extension.

niemela:
Closing after discussion on the December 9 meeting of the working group.

The plan is to add a list of removed intervals to the `/state` endpoint to allow downstream systems to detect removed intervals.

PR #60, opened 2021-10-25
Add multiple versions in repo

niemela:
Add explicit copies of all versions of the spec so that they can all be viewed on the webpage, and we can work on them in parallel.

* draft - the current version.
* 2021-11-draft - currently an exact copy of the current version.
* 2020-03 - from branch release-2020-03.

Depends on #59


thijskh:
So is there now any way to link to the "current" version of the spec?

Also, couldn't git branches have solved this problem?

eldering:
> So is there now any way to link to the "current" version of the spec? My links are now broken and I'm not sure what to replace them with.
> 
> Also, couldn't git branches have solved this problem?

See https://github.com/icpc/ccs-specs/pull/71 for a fix that puts different versions back into separate branches.

PR #66, opened 2021-11-08
Accounts and people

deboer-tim:
We have one remaining tsv file (accounts.tsv) and the CMS has additional data for staff that we sometimes use. This change adds accounts and generalizes team-members to allow other staff or be linked to accounts:
- Generalize `team-members` to `people` so that it can include staff. Adds a staff role and an optional title.
- Add accounts with several optional attributes.
- Add an optional `account_id` to submissions and clarifications so you can see which account made the change (in case there are multiple accounts per team, or an account per person).

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1134,7 +1134,7 @@ JSON elements of problem objects:
 | id                | ID      | yes       | no        | Identifier of the account.
 | username          | string  | yes       | no        | The account username.
 | password          | string  | no        | yes       | The account password.
-| type              | string  | no        | yes       | The type of account, one of `team`, `judge`, `admin`, `analyst`, `staff`.
+| type              | string  | no        | yes       | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
 | ip                | string  | no        | yes       | IP address associated with this account, used for auto-login.
 | team\_id          | ID      | no        | yes       | The team that this account is for. Required iff type is `team`.
 | people\_id        | ID      | no        | yes       | The person that this account is for, if the account is only for one person.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1093,9 +1093,9 @@ JSON elements of people objects:
 | :---------- | :------------- | :-------- | :-------- | :----------
 | id          | ID             | yes       | no        | Identifier of the person.
 | icpc\_id    | string         | no        | yes       | External identifier from ICPC CMS.
-| team\_id    | ID             | no        | no        | [Team](#teams) of this team member.
+| team\_id    | ID             | no        | yes       | [Team](#teams) of this person, if a team member.
 | name        | string         | yes       | no        | Name of the person.
-| title       | string         | no        | no        | Title of the person, e.g. "Technical director".
+| title       | string         | no        | yes       | Title of the person, e.g. "Technical director".
 | email       | string         | no        | yes       | Email of the person.
 | sex         | string         | no        | yes       | Either `male` or `female`, or possibly `null`.
 | role        | string         | yes       | no        | One of `contestant`, `coach`, or `staff`.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1093,7 +1093,7 @@ JSON elements of people objects:
 | :---------- | :------------- | :-------- | :-------- | :----------
 | id          | ID             | yes       | no        | Identifier of the person.
 | icpc\_id    | string         | no        | yes       | External identifier from ICPC CMS.
-| team\_id    | ID             | no        | yes       | [Team](#teams) of this person, if a team member.
+| team\_id    | ID             | no        | yes       | [Team](#teams) of this person. Required iff role is `team`.
 | name        | string         | yes       | no        | Name of the person.
 | title       | string         | no        | yes       | Title of the person, e.g. "Technical director".
 | email       | string         | no        | yes       | Email of the person.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1131,9 +1131,10 @@ JSON elements of problem objects:
 
 | Name              | Type    | Required? | Nullable? | Description
 | :---------------- | :------ | :-------- | :-------- | :----------
-| id                | ID      | yes       | no        | Identifier of the account, i.e. the username.
+| id                | ID      | yes       | no        | Identifier of the account.
+| user              | string  | yes       | no        | The account username.
 | password          | string  | no        | yes       | The account password.
-| type              | string  | yes       | no        | The type of account, one of `team`, `judge`, `admin`, `analyst`, `staff`.
+| type              | string  | no        | yes       | The type of account, one of `team`, `judge`, `admin`, `analyst`, `staff`.
 | ip                | string  | no        | yes       | IP address associated with this account, used for auto-login.
 | team\_id          | ID      | no        | yes       | The team that this account is for. Required iff type is `team`.
 | people\_id        | ID      | no        | yes       | The person that this account is for, if the account is only for one person.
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

draft/Contest_API.md
```diff
@@ -1084,44 +1084,95 @@ Returned data:
...
+| `/contests/<id>/accounts`      | application/json | yes       | JSON array of all accounts with elements as defined in the table below.
+| `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object of a single account with elements as defined in the table below.
+
+JSON elements of problem objects:
+
+| Name              | Type    | Required? | Nullable? | Description
+| :---------------- | :------ | :-------- | :-------- | :----------
+| id                | ID      | yes       | no        | Identifier of the account, i.e. the username.
+| password          | string  | no        | yes       | The account password.
+| type              | string  | yes       | no        | The type of account, one of `team`, `judge`, `admin`, `analyst`, `staff`.
```
nickygerritsen, line 1144: `staff` is new here, what does it mean?

nickygerritsen (approved):
I like this. It's simple enough and allows us to get rid of the last TSV

draft/Contest_API.md
deboer-tim: I added it to cover accounts for people like me that don't fit into the other categories (I am an admin for CDS, but not on the CCS). We could also leave null for these cases, or come up with another type.

draft/Contest_API.md
```diff
@@ -1084,44 +1084,95 @@ Returned data:
...
+| Endpoint                       | Mime-type        | Required? | Description
+| :----------------------------- | :--------------- | :-------- | :----------
+| `/contests/<id>/accounts`      | application/json | yes       | JSON array of all accounts with elements as defined in the table below.
+| `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object of a single account with elements as defined in the table below.
+
+JSON elements of problem objects:
+
+| Name              | Type    | Required? | Nullable? | Description
+| :---------------- | :------ | :-------- | :-------- | :----------
+| id                | ID      | yes       | no        | Identifier of the account, i.e. the username.
```
niemela, line 1142: Assumes that the allowed username space is a subset of our ID. Specifically, does not allow email as username, which is somewhat common.

draft/Contest_API.md
```diff
@@ -1084,44 +1084,96 @@ Returned data:
...
+| :----------------------------- | :--------------- | :-------- | :----------
+| `/contests/<id>/accounts`      | application/json | yes       | JSON array of all accounts with elements as defined in the table below.
+| `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object of a single account with elements as defined in the table below.
+
+JSON elements of problem objects:
+
+| Name              | Type    | Required? | Nullable? | Description
+| :---------------- | :------ | :-------- | :-------- | :----------
+| id                | ID      | yes       | no        | Identifier of the account.
+| user              | string  | yes       | no        | The account username.
```
niemela: I really don't like this. If it exists it should be `username` and not `user`, but I would rather have `id` of type `string` instead.

nickygerritsen: But then if we have an archive in CAF format and we later add `profile_picture`, where do we store it? Clearly we can't do `<contestid>/accounts/<id>/profile-picture.png` since the ID can be `/../../../../usr/bin/sudo`

nickygerritsen (dismissed):

draft/Contest_API.md
```diff
@@ -1084,44 +1084,96 @@ Returned data:
...
+| `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object of a single account with elements as defined in the table below.
+
+JSON elements of problem objects:
+
+| Name              | Type    | Required? | Nullable? | Description
+| :---------------- | :------ | :-------- | :-------- | :----------
+| id                | ID      | yes       | no        | Identifier of the account.
+| username          | string  | yes       | no        | The account username.
+| password          | string  | no        | yes       | The account password.
+| type              | string  | no        | yes       | The type of account, one of `team`, `judge`, `admin`, `analyst`, `staff`.
```
niemela, line 1145: Leave this open ended. I.e. allow systems to have other types of accounts.

niemela (changes requested):

draft/Contest_API.md
deboer-tim: Updated.

draft/Contest_API.md
deboer-tim: Separated id and username as per discussion.

nickygerritsen (approved):

niemela (approved):

draft/Contest_API.md
```diff
@@ -1084,44 +1084,96 @@ Returned data:
...
+JSON elements of problem objects:
+
+| Name              | Type    | Required? | Nullable? | Description
+| :---------------- | :------ | :-------- | :-------- | :----------
+| id                | ID      | yes       | no        | Identifier of the account.
+| username          | string  | yes       | no        | The account username.
+| password          | string  | no        | yes       | The account password.
+| type              | string  | no        | yes       | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
+| ip                | string  | no        | yes       | IP address associated with this account, used for auto-login.
+| team\_id          | ID      | no        | yes       | The team that this account is for. Required iff type is `team`.
```
eldering, line 1147: I think the required column here should contain "depends" to reflect the "Required iff type is `team`."

eldering (commented):
One tiny detail, looks fine otherwise.

draft/Contest_API.md
deboer-tim: Agreed, fixed (and hopefully rebased to latest).

eldering (approved):

niemela (approved):
Still needs a rebase (?).

deboer-tim:
> Still needs a rebase (?).

Appears I failed the first time. Tried again, got frustrated and went for the nuclear option. Hopefully ok now.

niemela:
The nuclear option is tried an true... https://xkcd.com/1597/

Contest_API.md
```diff
@@ -1076,44 +1076,96 @@ Returned data:
...
+| id          | ID             | yes       | no        | Identifier of the person.
 | icpc\_id    | string         | no        | yes       | External identifier from ICPC CMS.
-| team\_id    | ID             | yes       | no        | [Team](#teams) of this team member.
-| name        | string         | yes       | no        | Name of team member.
-| email       | string         | no        | yes       | Email of team member.
+| team\_id    | ID             | no        | yes       | [Team](#teams) of this person. Required iff role is `team`.
+| name        | string         | yes       | no        | Name of the person.
+| title       | string         | no        | yes       | Title of the person, e.g. "Technical director".
+| email       | string         | no        | yes       | Email of the person.
 | sex         | string         | no        | yes       | Either `male` or `female`, or possibly `null`.
```
niemela, line 1100: Minor comment: It's "nullable", do we really also need to say that it's "possible 'null'"?

Contest_API.md
```diff
@@ -1076,44 +1076,96 @@ Returned data:
...
-| role        | string         | yes       | no        | One of `contestant` or `coach`.
-| photo       | array of IMAGE | no        | yes       | Registration photo of the team member.
+| role        | string         | yes       | no        | One of `contestant`, `coach`, or `staff`.
```
niemela, line 1101: Similar to account type, maybe this should be allowed to be more open ended? Let's think about it for a possible future change?

niemela (approved):

eldering (approved):

PR #67, opened 2021-11-08
Update notification format & feed/webhook

deboer-tim:
As discussed on the call today, the event feed was receiving a redundant contest_id and the webhook was inconsistent in whether the callback would have one or more notification objects (and what the format was if more than one). This contains a few changes:
- Revert notification `endpoint` to `type`. This is one less change from the current spec, focuses on the contest object vs the REST API, and avoids some minor terminology confusion.
- Remove `contest_id` attribute from notification format. For event-feed this is already implied (you can only add a webhook to one contest), and for webhook each callback should have an array of notifications for only one contest per callback.
- Change webhook callback to an object that contains contest_id and an array of notification objects for that contest.
- Updated and added examples.

This change keeps the nice property that it would be trivial to convert between webhook and event feed or vice-versa, e.g. if you save every webhook notification array into a file called <contest_id>/event-feed.json as NDJSON, you essentially have a feed.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -346,19 +346,16 @@ etc.) or entire collection. The general format for events is:
 
 | Name        | Type            | Required? | Nullable? | Description
 | :---------- | :-------------- | :-------- | :-------- | :----------
-| type        | string          | yes       | yes       | The type of contest object that changed. Can be used for filtering.
+| type        | string          | yes       | no        | The type of contest object that changed. Can be used for filtering.
 | id          | string          | yes       | yes       | The id of the object that changed, or null for the entire collection/singleton.
 | data        | array or object | yes       | yes       | The updated value, i.e. what would be returned if calling the corresponding API endpoint at this time: an array, object, or null for deletions.
 
 Each event is a notification that an object or a collection has changed
 (and hence the contents of the corresponding endpoint) to `data`.
 
-If `type` and `id` are both non-null, then the object at
-`/contests/<contest_id>/<type>/<id>` now has the contents of `data`.
-If `id` is null, then the entire collection at `/contests/<contest_id>/<type>`
-now has the contents of `data`. If both are null, then the contest at
-`/contests/<contest_id>` now has the contents of `data`.
-
+If `type` is `contest`, then the contest at `/contests/<id>` now has the contents of `data`.
+If `id` is not null, then the object at `/contests/<contest_id>/<type>/<id>` now has the contents of `data`.
+If `id` is null, then the entire collection at `/contests/<contest_id>/<type>` now has the contents of `data`.
 
 #### Examples
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -322,7 +322,7 @@ are meant to ease extensibility.
 ### Notification format
 
 There are two mechanisms that clients can use to receive notifications
-of API updates (events), a webhook and a streaming HTTP feed. Both
+of API updates (events): a webhook and a streaming HTTP feed. Both
 mechanisms use the same payload format, but have different benefits,
 drawbacks, and ways to access. Webhooks are typically better for
 internet-scale, asynchronous processing, and disconnected systems; the
@@ -338,7 +338,7 @@ for these, since there will always be another event sent.
 
 The events are served as JSON objects, with every event corresponding to
 a change in a single object (submission, judgement, language, team,
-etc.) or full endpoint. The general format for events is:
+etc.) or entire collection. The general format for events is:
 
 ```json
 {"type": "<type>", "id": "<id>", "data": <JSON data for element> }
@@ -348,11 +348,17 @@ etc.) or full endpoint. The general format for events is:
 | :---------- | :-------------- | :-------- | :-------- | :----------
 | type        | string          | yes       | yes       | The type of contest object that changed. Can be used for filtering.
 | id          | string          | yes       | yes       | The id of the object that changed, or null for the entire collection/singleton.
-| data        | array or object | yes       | yes       | The object that would be returned if calling the corresponding API endpoint at this time, i.e. an array, object, or null for deletions.
+| data        | array or object | yes       | yes       | The updated value, i.e. what would be returned if calling the corresponding API endpoint at this time: an array, object, or null for deletions.
 
+Each event is a notification that an object or a collection has changed
+(and hence the contents of the corresponding endpoint) to `data`.
 
-The meaning of an event is to say that the contents at endpoint
-`/contests/<contest_id>/<type>/<id>` (or `/contests/<contest_id>/<type>` `/contests/<contest_id>`) now has the contents of `data`.
+If `type` and `id` are both non-null, then the object at
+`/contests/<contest_id>/<type>/<id>` now has the contents of `data`.
+If `id` is null, then the entire collection at `/contests/<contest_id>/<type>`
+now has the contents of `data`. If both are null, then the contest at
+`/contests/<contest_id>` now has the contents of `data`.
+
 
 #### Examples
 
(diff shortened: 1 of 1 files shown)


draft/Contest_API.md
```diff
@@ -349,26 +349,24 @@ a change in a single object (submission, judgement, language, team,
...
 
 | Name        | Type   | Required? | Nullable? | Description
 | :---------- | :----- | :-------- | :-------- | :----------
-| contest\_id | string | yes       | no        | The contest id.
-| endpoint    | string | yes       | yes       | The API endpoint, i.e. type of contest object above. Can be used for filtering.
-| id          | string | yes       | yes       | The id of the object that changed.
-| data        | object | yes       | yes       | The data is the object that would be returned if calling the corresponding API endpoint at this time, i.e. an object or null for deletions.
+| type        | string | yes       | yes       | The type of contest object that changed. Can be used for filtering.
+| id          | string | yes       | yes       | The id of the object that changed, or null for the entire collection/singleton.
+| data        | object | yes       | yes       | The object that would be returned if calling the corresponding API endpoint at this time, i.e. an object, or null for deletions.
```
nickygerritsen, line 359: Isn't `data` actually an object OR an array? (or null)

nickygerritsen (approved):

draft/Contest_API.md
deboer-tim: Yeah, I had almost updated that too. Did it now.

draft/Contest_API.md
```diff
@@ -349,26 +349,24 @@ a change in a single object (submission, judgement, language, team,
...
 
 | Name        | Type   | Required? | Nullable? | Description
 | :---------- | :----- | :-------- | :-------- | :----------
-| contest\_id | string | yes       | no        | The contest id.
-| endpoint    | string | yes       | yes       | The API endpoint, i.e. type of contest object above. Can be used for filtering.
-| id          | string | yes       | yes       | The id of the object that changed.
-| data        | object | yes       | yes       | The data is the object that would be returned if calling the corresponding API endpoint at this time, i.e. an object or null for deletions.
+| type        | string | yes       | yes       | The type of contest object that changed. Can be used for filtering.
+| id          | string | yes       | yes       | The id of the object that changed, or null for the entire collection/singleton.
+| data        | object | yes       | yes       | The object that would be returned if calling the corresponding API endpoint at this time, i.e. an array, object, or null for deletions.
```
nickygerritsen, line 359: I meant also the `Type` column. Should we say `object or array of objects`?

deboer-tim: Since it's referring to the type it seems natural enough to me that 'state' means the state changed and 'teams' means one or more teams changed (with details in the other attributes)... I can change it if you or others feel strongly about it though.

draft/Contest_API.md
```diff
@@ -349,26 +349,24 @@ a change in a single object (submission, judgement, language, team,
...
+{"type": "<type>", "id": "<id>", "data": <JSON data for element> }
 ```
 
 | Name        | Type   | Required? | Nullable? | Description
 | :---------- | :----- | :-------- | :-------- | :----------
-| contest\_id | string | yes       | no        | The contest id.
-| endpoint    | string | yes       | yes       | The API endpoint, i.e. type of contest object above. Can be used for filtering.
-| id          | string | yes       | yes       | The id of the object that changed.
-| data        | object | yes       | yes       | The data is the object that would be returned if calling the corresponding API endpoint at this time, i.e. an object or null for deletions.
+| type        | string | yes       | yes       | The type of contest object that changed. Can be used for filtering.
```
niemela, line 357: If we change from `endpoint` to `type` we need the type to endpoint mapping table (again).

draft/Contest_API.md
```diff
@@ -349,26 +349,24 @@ a change in a single object (submission, judgement, language, team,
...
+| id          | string | yes       | yes       | The id of the object that changed, or null for the entire collection/singleton.
+| data        | object | yes       | yes       | The object that would be returned if calling the corresponding API endpoint at this time, i.e. an array, object, or null for deletions.
 
 
 The meaning of an event is to say that the contents at endpoint
-`/contests/<contest_id>/<endpoint>/<id>` now has the contents of `data`.
+`/contests/<contest_id>/<type>/<id>` now has the contents of `data`.
```
niemela, line 363: Not technically true for `<type>` equal to `contest`.

draft/Contest_API.md
nickygerritsen: Or a place where we say that the type maps to the part after `/contest/<id>` except for `contest`?

draft/Contest_API.md
nickygerritsen: Still think the type of this `data` field should be `object or array`.

draft/Contest_API.md
deboer-tim: Updated as per discussion today.

draft/Contest_API.md
deboer-tim: Fixed.

draft/Contest_API.md
deboer-tim: Updated as per discussion today.

nickygerritsen (dismissed):

draft/Contest_API.md
```diff
@@ -346,29 +346,33 @@ for these, since there will always be another event sent.
...
+| data        | array or object | yes       | yes       | The updated value, i.e. what would be returned if calling the corresponding API endpoint at this time: an array, object, or null for deletions.
 
+Each event is a notification that an object or a collection has changed
+(and hence the contents of the corresponding endpoint) to `data`.
+
+If `type` and `id` are both non-null, then the object at
+`/contests/<contest_id>/<type>/<id>` now has the contents of `data`.
+If `id` is null, then the entire collection at `/contests/<contest_id>/<type>`
+now has the contents of `data`. If both are null, then the contest at
+`/contests/<contest_id>` now has the contents of `data`.
```
niemela, line 368: I thought we said that updates to `/contests/<contest_id>` should have type `contest`? This would imply that `type` can not be null.

Did I misunderstand?

deboer-tim: Nope, you're right. Updated and rebased.

Contest_API.md
```diff
@@ -338,29 +338,30 @@ for these, since there will always be another event sent.
...
+| type        | string          | yes       | no        | The type of contest object that changed. Can be used for filtering.
+| id          | string          | yes       | yes       | The id of the object that changed, or null for the entire collection/singleton.
+| data        | array or object | yes       | yes       | The updated value, i.e. what would be returned if calling the corresponding API endpoint at this time: an array, object, or null for deletions.
 
+Each event is a notification that an object or a collection has changed
+(and hence the contents of the corresponding endpoint) to `data`.
 
-The meaning of an event is to say that the contents at endpoint
-`/contests/<contest_id>/<endpoint>/<id>` now has the contents of `data`.
+If `type` is `contest`, then the contest at `/contests/<id>` now has the contents of `data`.
```
niemela, line 356: ...and `id` must be null.

Contest_API.md
```diff
@@ -338,29 +338,30 @@ for these, since there will always be another event sent.
...
+If `id` is not null, then the object at `/contests/<contest_id>/<type>/<id>` now has the contents of `data`.
+If `id` is null, then the entire collection at `/contests/<contest_id>/<type>` now has the contents of `data`.
```
niemela, line 358: Allowed values for `type` is not specified anywhere. The above three rows imply that `contest` or any of the endpoints under `/contests/<contest_id>/` are allowed, which is almost true.

I'm not sure if this must be fixed?

niemela (approved):
Some minor comments. Fix if you want to. (I'm leaning towards fixing it)

nickygerritsen (approved):

nickygerritsen (approved):

niemela (approved):

PR #71, opened 2021-11-21
Remove version directories

eldering:
This removes the directories with versions from the `master` branch and only keeps the current "draft" version.

The 2020-03 version was already contained in https://github.com/icpc/ccs-specs/tree/release-2020-03 (which I'd propose to rename to `2020-03`) and I've just pushed the 2021-11 draft to https://github.com/icpc/ccs-specs/tree/2021-11.

Publishing of the documentation to https://ccs-specs.icpc.io/ is done with the code in https://github.com/icpc/ccs-specs/tree/gh-pages


nickygerritsen (approved):

readme.md
```diff
@@ -1,30 +1,28 @@
...
 ## References
 
-Website: <https://ccs-specs.icpc.io>
-Github: <https://github.com/icpc/ccs-specs>
+* Website: <https://ccs-specs.icpc.io>
+* Github: <https://github.com/icpc/ccs-specs>
 
 The problem package format specification can be found at: <https://icpc.io/problem-package-format/>
-
-## Development notes
```
RagnarGrootKoerkamp, line 24: This is also the github readme. I'd prefer to keep this.

RagnarGrootKoerkamp (approved):
lgtm

readme.md
eldering: The issue is that this doesn't work out of the box anymore, since I moved the jekyll config to the common `gh-pages` branch to remove duplication and have a consistent config across all branches.

An option would be to link to the `gh-pages` branch readme and explain there how to build. I didn't test though whether that works out of the box on your local machine.

niemela:
> The 2020-03 version was already contained in https://github.com/icpc/ccs-specs/tree/release-2020-03 

Is the contents of the `release-2020-03` branch actually identical to the `2020-03` directory?

> (which I'd propose to rename to 2020-03)

That would remove the protection on the branch. The rule is currently that `master` and `release-*` are protected.

eldering:
> > The 2020-03 version was already contained in https://github.com/icpc/ccs-specs/tree/release-2020-03
> 
> Is the contents of the `release-2020-03` branch actually identical to the `2020-03` directory?

Yes, up to the changes in the last two commits on the branch https://github.com/icpc/ccs-specs/commits/release-2020-03

> > (which I'd propose to rename to 2020-03)
> 
> That would remove the protection on the branch. The rule is currently that `master` and `release-*` are protected.

Well, we could change the pattern easily, e.g. to `20[0-9][0-9]-[12][0-9]`. I think this has the advantage that it doesn't imply that a branch already contains an "officially released" version, but that it can still be in draft stage.

niemela:
> Well, we could change the pattern easily, e.g. to `20[0-9][0-9]-[12][0-9]`. I think this has the advantage that it doesn't imply that a branch already contains an "officially released" version, but that it can still be in draft stage.

Yeah, I really like that property. :+1:

niemela (approved):

niemela:
> Well, we could change the pattern easily, e.g. to `20[0-9][0-9]-[12][0-9]`.

I added this rule and will remove the old `release-*` rule when `release-2020-03` has been renamed.

nickygerritsen:
> > Well, we could change the pattern easily, e.g. to `20[0-9][0-9]-[12][0-9]`. I think this has the advantage that it doesn't imply that a branch already contains an "officially released" version, but that it can still be in draft stage.
> 
> Yeah, I really like that property. 👍

As far as I can find you can only use `*` and `?` in github branch patterns, so that is... meh

niemela:
> As far as I can find you can only use `*` and `?` in github branch patterns, so that is... meh

That's what I thought as well, but it does not seem to be the case anymore. (I believe this is the reason I used the `release-*` pattern before.

PR #74, opened 2022-02-15
Add problem package and statement to problems endpoint

niemela:
Add the problem package and statement to the problems endpoint.

Also, merge the file reference types (IMAGE, VIDEO, ARCHIVE, STREAM) into a single file reference type (FILE). Note that FILE was already used, but not defined.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -251,22 +251,21 @@ absolute timestamps.
     as labels may be random characters and cannot be assumed to be
     suitable for display purposes.
   - File references
-    (types **`IMAGE`**, **`VIDEO`**, **`ARCHIVE`**, **`STREAM`**, and **`DOCUMENT`** in
-    the specification) are represented as a JSON object with elements as
-    defined below.
+    (type **`FILE`** in the specification) are represented as a JSON object 
+    with elements as defined below.
   - Arrays (type **`array of <type>`** in the specification) are built-in JSON 
     arrays of some type defined above.
 
 Element for file reference objects:
 
-| Name     | Type    | Nullable?          | Description
-| -------- | ------- | ------------------ | -----------
-| href     | string  | yes                | URL where the resource can be found. Relative URLs are relative to the `baseurl`. Must point to a file of intended mime-type. Resource must be accessible using the exact same (possibly none) authentication as the call that returned this data.
-| filename | string  | no                 | POSIX compliant filename. Filenames must be unique within the endpoint object where they are used. I.e. an organization can have (multiple) `logo` and `country_flag` file references, they must all have a different filename, but different organizations may have files with the same filename.
-| hash     | string  | yes                | MD5 hash of the file referenced. 
-| mime     | string  | no                 | Mime type of resource.
-| width    | integer | no for **`IMAGE`** | Width of the image, video or stream in pixels. Should not be used for **`ARCHIVE`**.
-| height   | integer | no for **`IMAGE`** | Height of the image, video or stream in pixels. Should not be used for **`ARCHIVE`**.
+| Name     | Type    | Nullable? | Description
+| -------- | ------- | --------- | -----------
+| href     | string  | yes       | URL where the resource can be found. Relative URLs are relative to the `baseurl`. Must point to a file of intended mime-type. Resource must be accessible using the exact same (possibly none) authentication as the call that returned this data.
+| filename | string  | no        | POSIX compliant filename. Filenames must be unique within the endpoint object where they are used. I.e. an organization can have (multiple) `logo` and `country_flag` file references, they must all have a different filename, but different organizations may have files with the same filename.
+| hash     | string  | yes       | MD5 hash of the file referenced.
+| mime     | string  | no        | Mime type of resource.
+| width    | integer | depends   | Width of the image, video or stream in pixels. Required for files with mime type image/* and video/*.
+| height   | integer | depends   | Height of the image, video or stream in pixels. Required for files with mime type image/* and video/*.
 
 The `href` attributes may be [absolute or relative
 URLs](https://tools.ietf.org/html/rfc3986); relative URLs must be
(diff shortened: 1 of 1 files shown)


deboer-tim (commented):
I wasn't expecting the merge of file reference types. There is a slight drawback in that STREAMs are now FILEs, but that was inherent in the file reference type name anyway.

I am concerned with the new requirement for videos (team and reactions) & streams to have a width and height. We haven't done this before, and the latter requires us to either hardcode it or connect to every stream - would like to discuss on the next call.

niemela:
> I am concerned with the new requirement for videos (team and reactions) & streams to have a width and height. 

I would argue that this is not new. The old text said:
`Width of the image, video or stream in pixels. Should not be used for ARCHIVE.`

> We haven't done this before, and the latter requires us to either hardcode it or connect to every stream - would like to discuss on the next call.

I would have no trouble changing to:
`Width of the image. Required for files with mime type image/*.` (and same for height).
I only added the rest to *not* make a change.

deboer-tim:
> I would have no trouble changing to: `Width of the image. Required for files with mime type image/*.` (and same for height). I only added the rest to _not_ make a change.

Before it was "Nullable: no for **`IMAGE`**", now it's "Nullable: depends" and "Required for files with mime type image/* and video/*", which sounded like a change to me. I'd be good if we remove the "and video/*" as you suggested, or we can discuss.

niemela:
> I'd be good if we remove the "and video/*" as you suggested

Done.

Contest_API.md
```diff
@@ -900,6 +900,8 @@ JSON elements of problem objects:
 | time\_limit       | number  | no        | no        | Time limit in seconds per test data set (i.e. per single run). Should be an integer multiple of `0.001`.
 | test\_data\_count | integer | yes       | no        | Number of test data sets.
 | max_score         | number  | no        | no        | Maximum expected score, although teams may score higher in some cases. Typically used to indicate scoreboard cell color in scoring contests. Required iff contest:scoreboard_type is `score`.
+| problem_package   | array of FILE  | no | yes    | [Problem package](https://www.kattis.com/problem-package-format/). Expected mime type is application/zip. Only exactly one archive is allowed. Not expected to actually contain href for package during the contest, but used for configuration and archiving.
+| problem_statememt | array of FILE | no | yes    | Problem statemet. Expected mime type is application/pdf.
```
ghamerly, line 904:
```suggestion
| problem_statement | array of FILE | no | yes    | Problem statement. Expected mime type is application/pdf. 
```

ghamerly (dismissed):
Looks okay to me, with the one suggestion that I made (typo fixes).

Contest_API.md
```diff
@@ -900,6 +900,8 @@ JSON elements of problem objects:
 | time\_limit       | number  | no        | no        | Time limit in seconds per test data set (i.e. per single run). Should be an integer multiple of `0.001`.
 | test\_data\_count | integer | yes       | no        | Number of test data sets.
 | max_score         | number  | no        | no        | Maximum expected score, although teams may score higher in some cases. Typically used to indicate scoreboard cell color in scoring contests. Required iff contest:scoreboard_type is `score`.
+| problem_package   | array of FILE  | no | yes    | [Problem package](https://www.kattis.com/problem-package-format/). Expected mime type is application/zip. Only exactly one archive is allowed. Not expected to actually contain href for package during the contest, but used for configuration and archiving.
```
nickygerritsen, line 903: I get what you want to say with the sentence, but as a reader I got questions. Like "it's not expected during the contest, but how about before? Or after?" and "When **is** it expected?"

niemela: For configuration and archiving?

Do you have a suggestion for how to explain it better?

nickygerritsen: Not really. Maybe “for package in live systems” but that also is meh

eldering, line 903: Did we ever decide/clarify whether a problem package zip contains the files inside a directory of the problem name or in the root of the zip? I guess that's something that has to be clarified in the problem-package-format repo, but I think it's relevant here, as this is meant for potentially build automated tooling on top of that file format.

Contest_API.md
```diff
@@ -1019,20 +1021,20 @@ organizations (i.e. return any organization\_ids).
...
-| logo               | array of IMAGE | no        | yes       | Logo of the organization. A server must provide logos of size 56x56 and 160x160 but may provide other sizes as well.
+| Name               | Type          | Required? | Nullable? | Description
+| :----------------- | :------------ | :-------- | :-------- | :----------
+| id                 | ID            | yes       | no        | Identifier of the organization.
+| icpc\_id           | string        | no        | yes       | External identifier from ICPC CMS.
+| name               | string        | yes       | no        | Short display name of the organization.
+| formal\_name       | string        | no        | yes       | Full organization name if too long for normal display purposes.
+| country            | string        | no        | yes       | [ISO 3166-1 alpha-3 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) of the organization's country.
+| country_flag       | array of FILE | no        | yes       | Flag of the country. A server is recommended to provide flags of size around 56x56 and 160x160. Only allowed mime types are image/*.
```
eldering, line 1031: I'm assuming that here and below the intention is that this is still restricted to the 3 image mime types we support: png, jpg and svg, right? That's still mentioned above but is now a little less explicit, but I think that's ok.

Contest_API.md
niemela: I think what I wrote is better.

niemela: > Did we ever decide/clarify whether a problem package zip contains the files inside a directory of the problem name or in the root of the zip?

I don't remember.

I think we should say that the directory must be included.

> I guess that's something that has to be clarified in the problem-package-format repo, 

Yes.

> but I think it's relevant here, as this is meant for potentially build automated tooling on top of that file format.

Sure it's relevant, but the correct place to document all the details about the problem package format is in the problem package format specification. 

In summary, I don't think we should change the current wording here, but we should update the problem package format to clarify this.

Contest_API.md
niemela: > I'm assuming that here and below the intention is that this is still restricted to the 3 image mime types we support: png, jpg and svg, right? 

Yes.

> That's still mentioned above but is now a little less explicit, but I think that's ok.

I agree.

nickygerritsen (approved):

ghamerly (approved):

PR #75, opened 2022-02-16
Simplify Archive Format description

niemela:
Now when the API and Archive are in perfect sync, the description can be made *much* shorter.

Also, added some example uses cases with example file path listings.

In the future we need to add more examples and we should also create actual example archive package examples, and not only file path listings.


Contest_Archive_Format.md
```diff
@@ -16,708 +17,151 @@ disk, including:
...
+contest) in a single archive. One example of this would be a contest that was
+running with a primary and [shadow](ccs_system_requirements#shadow-mode) CCS.
+Typically in such cases, most of the data will be identical (or at least the
+differences irrelevant), so only the data that differs (in relevant ways)
+needs to be stored. If there is an additonal source for `<endpoint>` it is
+stored as above, but wherever `<endpoint>` is used in the path, instead use
+`<endpoint>_<system>` where `<system>` is a unique name for the additional
+source.
+
+## Exampole uses
```
nickygerritsen, line 70: Typo

Contest_Archive_Format.md
```diff
@@ -16,708 +17,151 @@ disk, including:
...
+
+### Registration upload
+
+Used for uploading local registration data to a central registration system
+(such as the ICPC CMS).
+
+Required endpoints:
+- teams
+- people
+
```
nickygerritsen, line 134: Does this not require (or optional) org info?

Contest_Archive_Format.md
```diff
@@ -16,708 +17,151 @@ disk, including:
...
   - As a base for offline analysis
 
-This standard lays out the relative location and format of each type of
-contest-related information when reading or writing to disk. The top
-level structure is inspired by the [Contest API](contest_api) structure.
-
 ## Archive contents
 
 The package consists of a single directory containing files as described
 below, or alternatively, a ZIP compressed archive of the same files. It is
```
eldering, line 23: Similar to the problem package format: good to clarify that the zip file also includes that directory (or not?).

Contest_Archive_Format.md
```diff
@@ -16,708 +17,151 @@ disk, including:
...
-
-```json
-{
-  "id": "asia",
-  "icpc_id": "7593",
-  "name": "Asia"
-}
+A package contains information regarding a single contest (corresponding to
+the `/contests/<id>/*` endpoints of the API). The API can contain information
+for several contests, to store all information a pacakge per contest would be
```
eldering, line 29: Typo `pacakge`

Contest_Archive_Format.md
niemela: Yes... it annoyingly does. Thanks.

My thinking was that the upload should not support adding new universities to the ICPC database (to keep things simple), but since the only university reference in `teams.json` is a `organization_id` (which could just be an arbitrary string) you do need `organizations.json`, if only to map `organization_id` to `icpc_id`.

I have added this.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Clarified that the directory *is* included.

Contest_Archive_Format.md
niemela: Fixed.

nickygerritsen (dismissed):

niemela:
I added the version information now that #76 has been merged.

nickygerritsen (approved):

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
-
-```json
-{
-  "id": "asia",
-  "icpc_id": "7593",
-  "name": "Asia"
-}
+A package contains information regarding a single contest (corresponding to
+the `/contests/<id>/*` endpoints of the API). The API can contain information
+for several contests, to store all information a package per contest would be
```
ghamerly, line 29:
```suggestion
for several contests, to store the information for multiple contests a package per contest would be
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+needed.
+
+Information in the API is always either in JSON format, [NDJSON]
+(contest_api#event-feed) format, or linked using a [file reference]
+(contest_api#json-attribute-types) JSON object.
```
ghamerly, line 34: I think the line breaks between [NDJSON] and (contest_api#event-feed) and similar for [file reference] and (contest_api#json-attribute-types) are causing the markdown to render incorrectly.

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+
+- The JSON returned from the endpoint `/` is stored as
+  `api.json`.
+- The JSON returned from the endpoint `/contests/<id>` is stored as
+  `contest.json`. (Notice the singular form).
+- The JSON returned from the endpoint `/contests/<id>/<endpoint>` is stored as
+  `<endpoint>.json`.
+- The NDJSON returned from the endpoint `/contests/<id>/<endpoint>` is stored as
+  `<endpoint>.ndjson`. (The only such endpoint is `event-feed`)
```
ghamerly, line 43:
```suggestion
  `<endpoint>.ndjson`. (The only such endpoint is `event-feed`.) 
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+- The JSON returned from the endpoint `/contests/<id>` is stored as
+  `contest.json`. (Notice the singular form).
+- The JSON returned from the endpoint `/contests/<id>/<endpoint>` is stored as
+  `<endpoint>.json`.
+- The NDJSON returned from the endpoint `/contests/<id>/<endpoint>` is stored as
+  `<endpoint>.ndjson`. (The only such endpoint is `event-feed`) 
+
+
+Files referenced to in `contest.json` are stored as `contest/<filename>`, and
+files referenced to in `<endpoint>.json` as `<endpoint>/<id>/<filename>`,
```
ghamerly, line 47:
```suggestion
files referenced in `<endpoint>.json` are stored as `<endpoint>/<id>/<filename>`,
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+  `api.json`.
+- The JSON returned from the endpoint `/contests/<id>` is stored as
+  `contest.json`. (Notice the singular form).
+- The JSON returned from the endpoint `/contests/<id>/<endpoint>` is stored as
+  `<endpoint>.json`.
+- The NDJSON returned from the endpoint `/contests/<id>/<endpoint>` is stored as
+  `<endpoint>.ndjson`. (The only such endpoint is `event-feed`) 
+
+
+Files referenced to in `contest.json` are stored as `contest/<filename>`, and
```
ghamerly, line 46:
```suggestion
Files referenced in `contest.json` are stored as `contest/<filename>`, and
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+files referenced to in `<endpoint>.json` as `<endpoint>/<id>/<filename>`,
+where:
+- `<id>` is the ID of the endpoint object the reference is in.
+- `<filename>` is the filename specified in the file reference object.
+
+Note that the API specification requires that filenames are unique within
+endpoint objects, so this is always possible.
+
+The href in the file reference may not be valid and should be ignored, but it
+does not have to be removed.
```
ghamerly, line 56: This sentence seems to jump out as needing some explanation or context. Could we explain when or why this might happen?

Also, I'm not sure what it means. Would it be valid to rewrite it as: "It is possible that the `href` in the file reference is not valid, and if it is invalid it should be ignored (but need not be removed)."

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+
+Optionally one could create a Shallow Archive by not storing the files, but
+then the hrefs must be valid. This could be useful in some cases where the
+size of the archive matters.
+
+In some cases it could make sense to merge multiple API sources (of the same
+contest) in a single archive. One example of this would be a contest that was
+running with a primary and [shadow](ccs_system_requirements#shadow-mode) CCS.
+Typically in such cases, most of the data will be identical (or at least the
+differences irrelevant), so only the data that differs (in relevant ways)
```
ghamerly, line 66:
```suggestion
differences should be irrelevant), so only the data that differs (in relevant ways)
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
+...
+teams.json
+teams/team-001/photo.jpg
+teams/team-002/photo.jpg
+teams/team-003/photo.jpg
+...
+people.json
+people/john-smith/photo.jpg
+people/jane-doe/photo.jpg
+....
```
ghamerly, line 126:
```suggestion
...
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,166 @@ disk, including:
...
```
ghamerly, line 159:
```suggestion
...
```

ghamerly (changes requested):
A few minor comments / suggestions. The only real issue I would like clarified is the "invalid href".

Contest_Archive_Format.md
niemela: The href might point to a location on a temporary local network, and thus not be available after the contest.

Contest_Archive_Format.md
niemela: I don't think "should" is right here. Could "would" work? Or could we really not keep it as is?

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: I made an attempt of a better explanation...

Contest_Archive_Format.md
```diff
@@ -16,708 +17,170 @@ disk, including:
...
+Files referenced in `contest.json` are stored as `contest/<filename>`, and
+files referenced in `<endpoint>.json` are stored as
+`<endpoint>/<id>/<filename>`, where:
+- `<id>` is the ID of the endpoint object the reference is in.
+- `<filename>` is the filename specified in the file reference object.
+
+Note that the API specification requires that filenames are unique within
+endpoint objects, so this is always possible.
+
+It is not required that URLs specified in the href is always valid.
```
ghamerly, line 55:
```suggestion
It is not required that a URL specified in an href is always valid.
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,170 @@ disk, including:
...
+Specifically, in many cases the contest is running in a local network that is
+taken down after the contest, and this the URL would definitely not be
```
ghamerly, line 57:
```suggestion
taken down after the contest, and in this case the URL would definitely not still be
```

Contest_Archive_Format.md
```diff
@@ -16,708 +17,170 @@ disk, including:
...
+working. To keep the archiving process as simple as possible, stale URLs do
+not have to be removed from the data, but due to this the URLs should be
+ignored and the file stored as specified above should be used instead.
+
+Optionally one could create a Shallow Archive by not storing the files Then
```
ghamerly, line 62:
```suggestion
Optionally one could create a Shallow Archive by not storing the files, in which case the
```

Contest_Archive_Format.md
ghamerly: The phrase is missing a verb. How about "are" instead of "should be"?

ghamerly (changes requested):

Contest_Archive_Format.md
niemela: In that case:
```
most of the data is identical (or at least the differences are irrelevant),
```

I.e. change "will be" to "is" and add the "are". 

This feels better than the original to me. Would it work?

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: Fixed.

Contest_Archive_Format.md
niemela: I did that.

ghamerly (approved):

nickygerritsen (approved):

eldering (approved):

PR #76, opened 2022-02-17
Add version information to the API root

niemela:
Add version name and link to the root of the API.

(Note, we need to add this to the Archive Format as well, but waiting for #75 to be merged.)


Contest_API.md
```diff
@@ -538,6 +538,38 @@ meaning that all conditions must be met (they are logically `AND`ed).
 Note that filtering on any other field, including fields with the type
 array of ID, does not have to be supported.
 
+### API version
```
nickygerritsen, line 541: If we add more fields in the future the name `version` might not be ideal. But we can change it at that point, that doesn't impact backwards compatibility.

nickygerritsen (dismissed):

deboer-tim (approved):
Nit: should be "Will be of the form" not "Will be on the form". Maybe a 'must' too?

PR #78, opened 2022-02-25
Consistent terminology of 'object' and 'property'

deboer-tim:
The spec used a wide mix of terms to mean the same things, sometime overlapping. As discussed on yesterday's call, I've done a pass to make it consistent:
- attribute -> property (except in one case where it was referring to an HTTP query parameter.
- field -> property.
- element -> depending on the use this changed to object or property.
- JSON list -> JSON array, containing items.
- JSON elements -> Properties. Could have left 'JSON properties', but I thought this was simpler and obvious enough from the context.

Also made two minor tweaks for consistency.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -58,13 +58,13 @@ requested, e.g. at the URL path
 
 ` GET https://example.com/api/contests/wf14/teams`
 
-while an element with specific ID is requested as
+while an object with a specific ID is requested as
 
 ` GET https://example.com/api/contests/wf14/teams/10`
 
-A collection is always returned as a JSON list of objects. Every object
-in the list represents a single element (and always includes the ID).
-When requesting a single element the exact same object is returned. E.g.
+A collection is always returned as an array of JSON objects. Every item
+in the array is a single contest object (and always includes the ID).
+When requesting a single object the exact same object is returned. E.g.
 the URL path
 
 `GET baseurl/collection`
@@ -72,8 +72,8 @@ the URL path
 returns
 
 ```json
-[ { "id":<id1>, <element specific data for id1> },
-  { "id":<id2>, <element specific data for id2> },
+[ { "id":<id1>, <object specific data for id1> },
+  { "id":<id2>, <object specific data for id2> },
      ...
 ]
 ```
@@ -84,7 +84,7 @@ while the URL path
 
 returns
 
-`{ "id":<id1>, <element specific data for id1> }`
+`{ "id":<id1>, <object specific data for id1> }`
 
 ### HTTP headers
 
@@ -110,30 +110,30 @@ their expected behavior, if implemented.
 
   - `GET`
     Read data. This method is idempotent and does not modify any data.
-    It can be used to request a whole collection or a specific element.
+    It can be used to request a whole collection or a specific object.
 
   - `POST`
-    Create a new element. This can only be called on a collection
-    endpoint, and the `id` attribute may not be specified as it is up
+    Create a new object. This can only be called on a collection
+    endpoint, and the `id` property may not be specified as it is up
     to the server to assign one.
     If successful the response will contain a `Location` header
-    pointing to the newly created element.
+    pointing to the newly created object.
 
   - `PUT`
-    Creates or replaces a specific element. This method is idempotent, can only
-    be called on a specific element, and replaces its contents with the
+    Creates or replaces a specific object. This method is idempotent, can only
+    be called on a specific object, and replaces its contents with the
     data provided. The payload data must be complete, i.e. the `id` is
     required and no partial updates are allowed.
 
   - `PATCH`
-    Updates/modifies a specific element. This method is idempotent,
-    can only be called on a specific element, and replaces the given
-    attributes with the data provided. For example
+    Updates/modifies a specific object. This method is idempotent,
+    can only be called on a specific object, and replaces the given
+    properties with the data provided. For example
     `PATCH https://example.com/api/contests/wf14/teams/10`
     with JSON contents `{"name":"Our cool new team name"}`.
 
   - `DELETE`
-    Delete a specific element. Idempotent, but will return a `404` error
+    Delete a specific object. Idempotent, but will return a `404` error
     code when repeated. Any provided data is ignored, and there is no response body if successful.
     Example: `DELETE https://example.com/api/contests/wf14/teams/8`.
     Note that deletes must keep [referential integrity](#referential-integrity) intact.
@@ -143,24 +143,24 @@ their expected behavior, if implemented.
 Standard [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
 returned to indicate success or failure. If successful DELETE will have no response body,
 GET on a collection will return the collection, and every other method will contain the
-current (updated) state of the element.
+current (updated) state of the object.
 
-If a POST, PUT, PATCH would cause any of the following issues it must fail, in addition to any endpoint or element-specific requirements:
+If a POST, PUT, PATCH would cause any of the following issues it must fail, in addition to any endpoint or object-specific requirements:
 
 * A PATCH on an `id` that doesn't exist. Will return a 404 error code.
 * A PUT or PATCH containing an id that does not match the URL. Will return a 409 error code.
-* A missing required attribute.
-* An attribute that must not be provided is provided.
-* An attribute type that is incorrect or otherwise invalid (e.g. non-nullable attribute set to null).
-* A reference to another element is invalid (to maintain [referential integrity](#referential-integrity)).
+* A missing required property.
+* A property that must not be provided is provided.
+* A property type that is incorrect or otherwise invalid (e.g. non-nullable property set to null).
+* A reference to another object is invalid (to maintain [referential integrity](#referential-integrity)).
 
-In addition to any endpoint or element-specific requirements, DELETE must fail
-if the element `id` doesn't exist, and return a 404 error code. 
-If the element being deleted is referenced by another element, the server must either
+In addition to any endpoint or object-specific requirements, DELETE must fail
+if the object `id` doesn't exist, and return a 404 error code. 
+If the object being deleted is referenced by another object, the server must either
 fail or implement a cascading delete (to maintain [referential integrity](#referential-integrity))
 
 When there is a failure using any method the response message body
-must include a JSON element that contains the attributes 'code' (a number,
+must include a JSON object that contains the properties 'code' (a number,
 identical to the HTTP status code returned) and 'message' (a string) with
 further information suitable for the user making the request, as per the
 following example:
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

Contest_API.md
```diff
@@ -58,22 +58,22 @@ requested, e.g. at the URL path
...
-while an element with specific ID is requested as
+while an object with a specific ID is requested as
 
 ` GET https://example.com/api/contests/wf14/teams/10`
 
-A collection is always returned as a JSON list of objects. Every object
-in the list represents a single element (and always includes the ID).
-When requesting a single element the exact same object is returned. E.g.
+A collection is always returned as an array of JSON objects. Every item
+in the array is a single contest object (and always includes the ID).
```
niemela, line 66: I understand that "contest object" is used here in a general way (i.e. one of the objects related to the contest), but a contest object could more specifically be the object returned from the contest endpoint. (This is exactly what we call it on line 584).

I suggest dropping "contest".

Contest_API.md
```diff
@@ -497,45 +497,45 @@ NDJSON format.
 
 In the tables below, the columns are:
 
-  - Name: Attribute name; object sub-attributes are indicated as
-    `object.attribute`.
-  - Type: Data type of the attribute; one of the 
-    [types listed above](#json-attribute-types).
-  - Required?: Whether this is a required attribute that **must** be
+  - Name: Property name; object sub-properties are indicated as
+    `object.property`.
```
niemela, line 501: The object would be the submission or the judgement, so this should really be `property.subproperty` not `object.subproperty`.

Also, is it clear what a "sub-property" is?

Contest_API.md
```diff
@@ -58,22 +58,22 @@ requested, e.g. at the URL path
...
 
 `GET baseurl/collection`
 
 returns
 
 ```json
-[ { "id":<id1>, <element specific data for id1> },
-  { "id":<id2>, <element specific data for id2> },
+[ { "id":<id1>, <object specific data for id1> },
+  { "id":<id2>, <object specific data for id2> },
```
niemela, line 76: I find "object specific" a bit confusing here. How about:

```
[ { "id":<id1>, <properties for object with id=id1> },
  { "id":<id2>, <properties for object with id=id2> },
```
?

Contest_API.md
```diff
@@ -84,7 +84,7 @@ while the URL path
 
 returns
 
-`{ "id":<id1>, <element specific data for id1> }`
+`{ "id":<id1>, <object specific data for id1> }`
```
niemela, line 87: Same as above.

Contest_API.md
```diff
@@ -143,24 +143,24 @@ their expected behavior, if implemented.
 Standard [HTTP status codes](https://en.wikipedia.org/wiki/List_of_HTTP_status_codes) are
 returned to indicate success or failure. If successful DELETE will have no response body,
 GET on a collection will return the collection, and every other method will contain the
-current (updated) state of the element.
+current (updated) state of the object.
 
-If a POST, PUT, PATCH would cause any of the following issues it must fail, in addition to any endpoint or element-specific requirements:
+If a POST, PUT, PATCH would cause any of the following issues it must fail, in addition to any endpoint or object-specific requirements:
```
niemela, line 148: Would "endpoint-specific" be better?

Contest_API.md
```diff
@@ -546,9 +546,9 @@ The following endpoint is associated with API version:
 
 | Endpoint | Mime-type        | Required? | Description
 | :------- | :--------------- | :-------- | :----------
-| `/`      | application/json | yes       | JSON object with version information as defined in the table below.
+| `/`      | application/json | yes       | JSON object with properties as defined in the table below.
```
niemela, line 549: All other endpoint descriptions are on the form "JSON object of X with properties as defined in the table below". The X here should be "version information".

That said, the "JSON object of X" would sound a bit strange here, but it also sounds a bit strange everywhere, doesn't it?

Contest_API.md
```diff
@@ -1561,10 +1562,10 @@ The following endpoints are associated with clarification messages:
 
 | Endpoint                             | Mime-type        | Required? | Description
 | :----------------------------------- | :--------------- | :-------- | :----------
-| `/contests/<id>/clarifications`      | application/json | yes       | JSON array of all clarification messages with elements as defined in the table below.
-| `/contests/<id>/clarifications/<id>` | application/json | yes       | JSON object of a single clarification message with elements as defined in the table below.
+| `/contests/<id>/clarifications`      | application/json | yes       | JSON array of all clarifications with properties as defined in the table below.
+| `/contests/<id>/clarifications/<id>` | application/json | yes       | JSON object of a single clarification with properties as defined in the table below.
 
-JSON elements of clarification message objects:
+Properties of clarification objects:
```
niemela, line 1568: I don't agree with the dropping of "message".

A clarification request is not a clarification, but both are (arguably) clarification messages.

Contest_API.md
```diff
@@ -1937,9 +1938,9 @@ JSON elements of scoreboard row objects:
 | score.total\_time | integer | depends   | no        | Total penalty time accrued by the team. Required iff contest:scoreboard_type is `pass-fail`.
 | score.score       | number  | depends   | no        | Total score of problems by the team. Required iff contest:scoreboard_type is `score`.
 | score.time        | integer | no        | no        | Time of last score improvement used for tiebreaking purposes.
-| problems          | array of problem data objects | yes       | no        | JSON array of problems with scoring data, see below for the specification of each element.
+| problems          | array of problem data objects | yes       | no        | JSON array of problems with scoring data, see below for the specification of each problem.
```
niemela, line 1941: Shouldn't "problem" be "object"?

Contest_API.md
eldering: It already says that: "endpoint or object-specific requirements"

Contest_API.md
```diff
@@ -187,31 +187,31 @@ additional roles may be supported for specific uses:
...
 not have access to judgements and runs from submissions made after the
 contest freeze time.
 
 ### Referential integrity
 
-Some attributes in elements are references to IDs of other elements.
-When such an attribute has a non-`null` value, then the referenced
-element must exist. That is, the full set of data exposed by the API
+Some properties in an object are references to IDs of other objects.
+When such an property has a non-`null` value, then the referenced
```
eldering, line 199: Tiny typo:
```suggestion
When such a property has a non-`null` value, then the referenced
```

Contest_API.md
```diff
@@ -308,16 +308,16 @@ with JSON data
...
-  - Clients should accept extra attributes in endpoints, that are not
+  - Clients should accept extra properties in endpoints, that are not
     specified here.
   - Servers should not expect clients to recognize more than the basic,
     required specification.
-  - In this specification and extensions, an attribute with value `null`
+  - In this specification and extensions, a property with value `null`
     may be left out by the server (i.e. not be present). A client must
-    treat an attribute with value `null` equivalently as that attribute
+    treat an property with value `null` equivalently as that property
```
eldering, line 320:
```suggestion
    treat a property with value `null` equivalently as that property
```

Contest_API.md
eldering: We could mention that we're essentially using JSONPath notation. That might also clarify the meaning of sub-property.

Contest_API.md
```diff
@@ -1598,12 +1599,12 @@ endpoint returns with a `GET` request with the following exceptions:
...
   proxy. See notes under the submission interface for more detail. In cases where
-  these attributes are not allowed the server will respond with a 4xx error code.
+  these properties are not allowed the server will respond with a 4xx error code.
 
 The request must fail with a 4xx error code if any of the following happens:
 
-* A required attribute is missing.
-* An attribute that must not be provided is provided.
+* A required property is missing.
+* An property that must not be provided is provided.
```
eldering, line 1607:
```suggestion
* A property that must not be provided is provided.
```

Contest_API.md
niemela: Is there some good documentation for JSONPath notation we could link to?

Contest_API.md
niemela: Indeed.

Contest_API.md
deboer-tim: What do you think about "JSON object representing a single X"?

deboer-tim: Sorry, I see you gave a thumbs up in our slack :)

Contest_API.md
deboer-tim: Fixed

Contest_API.md
deboer-tim: Based on slack with Fredrik I've changed these to `{ "id":<id1>, <other properties for object id1> }`.

Contest_API.md
deboer-tim: Fixed as above.

Contest_API.md
deboer-tim: Fixed in new PR.

Contest_API.md
```diff
@@ -143,24 +143,24 @@ their expected behavior, if implemented.
...
 * A PUT or PATCH containing an id that does not match the URL. Will return a 409 error code.
-* A missing required attribute.
-* An attribute that must not be provided is provided.
-* An attribute type that is incorrect or otherwise invalid (e.g. non-nullable attribute set to null).
-* A reference to another element is invalid (to maintain [referential integrity](#referential-integrity)).
-
-In addition to any endpoint or element-specific requirements, DELETE must fail
-if the element `id` doesn't exist, and return a 404 error code. 
-If the element being deleted is referenced by another element, the server must either
+* A missing required property.
```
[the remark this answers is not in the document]
deboer-tim: Agreed, fixed in new PR.

Contest_API.md
deboer-tim: I was on the fence because the object is called 'clarification', not 'clarification-message', but I've reverted this in the next commit.

Contest_API.md
deboer-tim: Fixed in 2nd commit.

Contest_API.md
deboer-tim: Fixed in 2nd commit.

Contest_API.md
deboer-tim: Fixed in 2nd commit.

Contest_API.md
deboer-tim: I fixed the text to read property.subproperty in the 2nd commit, let me know the text/link if you want to include in this PR, otherwise we can add the JSONPath reference separately.

eldering (approved):

Contest_API.md
eldering: Since this is not exactly JSONPath (that always starts with `$`) and I couldn't find a nice reference, I'm happy to leave this as is.

niemela: I think this is good for now.

niemela (approved):

PR #80, opened 2022-03-03
Fix typos found by codespell

vmcj:
Not 100% sure about all of those.

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -963,9 +963,9 @@ problem.
 
 ### Contest API
 
-It is a requirement that the CCS provide a compliant implmentation of
+It is a requirement that the CCS provide a compliant implementation of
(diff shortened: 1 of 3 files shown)


deboer-tim (approved):

Contest_Control_System_Requirements.md
```diff
@@ -80,7 +80,7 @@ document.
 
 ### Licensing
 
-The CCS must either be freely useable, or must be accompanied by a
+The CCS must either be freely usable, or must be accompanied by a
```
niemela, line 83: I believe this is British vs American English, isn't it?

What are we using?

niemela (approved):

Contest_Control_System_Requirements.md
eldering: FYI: we are using American English, see 8e433ffb0954fa42fbb606399d046e513304233f.

PR #81, opened 2022-03-04
Capabilities and Account endpoint

deboer-tim:
I looked at several REST APIs, and the pattern is often /users (for all users), /users/x for a specific user, and /user to get the current user. In more than one case /users/x and /user had different properties.

Following this same pattern, I'm proposing /account to get the currently logged in account. The properties are identical to /accounts where they overlap (id, username, team/people id), but it does not include all of the same properties (e.g. password, IP) and has a couple new ones: capabilities, and a list of types that tells you which types and properties for each you may see. The types list is shared with the notification format.

I've added a section to describe what capabilities are, but not attempt to define any of them. I'd like to focus on the accounts endpoint and capabilities description first, then use separate PR(s) to define the capabilities.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1260,7 +1260,7 @@ Properties of account objects:
 | :----------- | :-------------------- | :-------- | :-------- | :----------
 | id           | ID                    | yes       | no        | Identifier of the account used to make this request.
 | username     | string                | yes       | no        | The account username.
-| team\_id     | ID                    | depends   | yes       | The team that this account is for. Required iff type is `team`.
+| team\_id     | ID                    | no        | yes       | The team that this account is for, if the account is for a team, or one team member.
 | people\_id   | ID                    | no        | yes       | The person that this account is for, if the account is only for one person.
 | capabilities | array of string       | no        | yes       | An array of [capabilities](#capabilities) that the current account has.
 | types        | array of type objects | yes       | no        | An array of type objects, as described below.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -322,6 +322,24 @@ are meant to ease extensibility.
     treat a property with value `null` equivalently as that property
     not being present.
 
+### Capabilities
+
+Instead of strict role-based access control, this API defines several
+capabilities that define the behaviours that clients can expect and
+actions they can perform. For instance, a team account will typically
+have access to a "team_submit" capability that allows a team to perform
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1226,6 +1244,70 @@ Returned data:
...
+| :----------------------- | :--------------- | :-------- | :----------
+| `/contests/<id>/account` | application/json | yes       | JSON object representing a single account with properties as defined in the table below.
+
+Properties of account objects:
+
+| Name         | Type                  | Required? | Nullable? | Description
+| :----------- | :-------------------- | :-------- | :-------- | :----------
+| id           | ID                    | yes       | no        | Identifier of the account used to make this request.
+| username     | string                | yes       | no        | The account username.
+| team\_id     | ID                    | depends   | yes       | The team that this account is for. Required iff type is `team`.
```
nickygerritsen, line 1263: The `required iff` is not correct, since `type` is not a thing (anymore)

Contest_API.md
```diff
@@ -1226,6 +1244,70 @@ Returned data:
 {"id":"nicky","type":"admin"}
 ```
 
+### Account
+
+Provides information on which account the client represents within a given contest, which endpoints properties are visible, and what [capabilities](#capabilities) (typically PUT, POST, PATCH, or DELETE actions) this account has access to or can perform.
```
nickygerritsen, line 1249: I think we still allow public access to the API (i.e. without an account). Maybe we should specify what happens in that case?

Contest_API.md
deboer-tim: Removed depends and updated description to match.

Contest_API.md
deboer-tim: I was about to just say that it returns 401 when not logged in, but even a public client might be helped by knowing which types and capabilities they have access to. What do you think about making the id and username not required and nullable for this case? Any other suggestion?

nickygerritsen: I was also leaning to making these fields nullable. Let's discuss tonight?

deboer-tim: k

deboer-tim:
As discussed on the call, we'll split this into /account and /access. It was easier to start a new branch, so I'm cancelling this PR/branch, PR #82 adds /account, and I'll create another for /access.

PR #82, opened 2022-03-07
Account endpoint

deboer-tim:
Replacing PR #81. This just adds a simple /account endpoint to match /accounts, allowing clients to know which account they are authenticated/logged in as.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1187,7 +1187,7 @@ The following endpoints are associated with accounts:
 | `/contests/<id>/accounts`      | application/json | yes       | JSON array of all accounts with properties as defined in the table below.
 | `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object representing a single account with properties as defined in the table below.
 
-Properties of problem objects:
+Properties of account objects:
 
 | Name              | Type    | Required? | Nullable? | Description
 | :---------------- | :------ | :-------- | :-------- | :----------
@@ -1224,6 +1224,31 @@ Returned data:
 
 ```json
 {"id":"nicky","type":"admin"}
+```
+
+### Account
+
+Provides information on which account the client represents within a given contest.
+
+The following endpoint is associated with an account:
+
+| Endpoint                 | Mime-type        | Required? | Description
+| :----------------------- | :--------------- | :-------- | :----------
+| `/contests/<id>/account` | application/json | yes       | JSON object representing a single account with properties as defined in the [Accounts](#accounts) table above.
+
+This endpoint exists in the API so that the clients can tell which account (and hence which person or team) they are logged in as. It is not
+expected to exist in a contest archive, and will fail with a 401 error code if the client is not authenticated.
+
+#### Examples
+
+Request:
+
+`GET https://example.com/api/contests/wf14/account`
+
+Returned data:
+
+```json
+{"id":"nicky","username":"Nicky"}
 ```
 
 ### Contest state


Contest_API.md
```diff
@@ -1226,6 +1226,31 @@ Returned data:
...
+Provides information on which account the client represents within a given contest.
+
+The following endpoint is associated with an account:
+
+| Endpoint                 | Mime-type        | Required? | Description
+| :----------------------- | :--------------- | :-------- | :----------
+| `/contests/<id>/account` | application/json | yes       | JSON object representing a single account with properties as defined in the [Accounts](#accounts) table above.
+
+This endpoint exists in the API so that the clients can tell which account (and hence which person or team) they are logged in as. It is not
+expected to exist in a contest archive, and will fail with a 401 error code if the client is not authenticated.
```
nickygerritsen, line 1240: 401 is the one where the server is expected to send a WWW-Authenticate header with a challenge, which the client should use to show a login prompt (see [here](https://www.webfx.com/web-development/glossary/http-status-codes/what-is-a-401-status-code/)). That doesn't seem like what we want? Maybe we want [403](https://www.webfx.com/web-development/glossary/http-status-codes/what-is-a-403-status-code/)?

niemela: That doesn't feel perfectly right either.

If this was a list we could return an empty list... but it isn't, and it shouldn't be.

Contest_API.md
```diff
@@ -1226,6 +1226,31 @@ Returned data:
...
 
+### Account
+
+Provides information on which account the client represents within a given contest.
+
+The following endpoint is associated with an account:
+
+| Endpoint                 | Mime-type        | Required? | Description
+| :----------------------- | :--------------- | :-------- | :----------
+| `/contests/<id>/account` | application/json | yes       | JSON object representing a single account with properties as defined in the [Accounts](#accounts) table above.
```
niemela, line 1237: I would suggest merging the "Account" section into the "Accounts" section. So, there would be three endpoints listed there.

niemela (changes requested):

Contest_API.md
nickygerritsen: Yeah I can't seem to find a better HTTP status code. But 401 is definitely not the correct one. Or if it is, we NEED to send this strange WWW-Authenticate header. Which seems wrong.

Contest_API.md

Contest_API.md
```diff
@@ -1226,6 +1230,16 @@ Returned data:
...
 ```
 
+Request:
+
+`GET https://example.com/api/contests/wf14/account`
+
+Returned data:
+
+```json
+{"id":"nicky","username":"Nicky"}
```
nickygerritsen, line 1240: oohhhhhh I only see this now. I feel very special.

Contest_API.md
deboer-tim: I was waffling between 401 and 403 initially, switched it to 403 now. Fredrik, Nicky and I can't find anything better, so please approve or provide another suggestion. :)

Contest_API.md
```diff
@@ -1178,16 +1178,17 @@ Returned data:
...
-The accounts used for access to the contest.
+The accounts used for accessing the contest, including the account representing the client.
 
 The following endpoints are associated with accounts:
 
 | Endpoint                       | Mime-type        | Required? | Description
 | :----------------------------- | :--------------- | :-------- | :----------
 | `/contests/<id>/accounts`      | application/json | yes       | JSON array of all accounts with properties as defined in the table below.
 | `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object representing a single account with properties as defined in the table below.
+| `/contests/<id>/account`       | application/json | yes       | JSON object representing a single account with properties as defined in the table below.
```
eldering, line 1189: I think it would be good to mention here that this is the account one is logged in with.

eldering (dismissed):

Contest_API.md
niemela: - 401 is definitely wrong.
- I don't think 403 is quite right either.
- 404 is kinda technically correct.
- An actual empty result (i.e. `""`) is also kinda technically correct.

If this was an array then returning an empty array (`"[]"`) would be nicer and clearer, but it isn't an array, and I don't think we should make it one just for this.

I think the empty result would be the best choice.

nickygerritsen: But that’s not valid json is it? Then it should return `null`.
What makes 403 not right? As far as I know it means “with your current credentials you do not have permission to view this entity” which kinda is true?

deboer-tim: I agree with Nicky that it'd have to return `null`. An empty array or "" are also valid JSON, but then you need to document that two different JSON types may be returned, which is awkward and harder for clients to deal with.
IMHO HTTP status code is the most straightforward to document and for clients to handle.

I have seen sites use 404 for things you need to log in to see, to avoid giving out any information to unauthenticated users. If the server didn't support unauthenticated access and is returning 404 for /contests then I'd assume it would do the same here, but if /contests exists for the user then 404 seems wrong to me; you know this endpoint exists.

An unauthenticated user asking what account they have is an edge case though. I disagree with `[]` or `""`, but if Fredrik feels strongly about 404 or `null` I'd rather change, merge, and discuss on the next call vs delaying getting the rest in. :)

niemela: > But that’s not valid json is it?

No, it seems it isn't. I don't know why I thought it would be. (Maybe because it totally should be 😄).

> Then it should return `null`.

But it also really shouldn't. `null` isn't a JSON object with properties as we have defined.

 Now, if everything is optional, then an empty *object* would be correct (i.e. `"{}"`).

> What makes 403 not right? As far as I know it means “with your current credentials you do not have permission to view this entity” which kinda is true?

Kinda, a bit, maybe. I would argue that 403 implies that there exists an object here, but you are not allowed to see it. But the object in fact does not exist, which would imply 404. If you *do* log i, you will get some *other* object, not this one. Which one depends on what you authenticate as.

So, let me update my stance:
- 401 is definitely wrong.
- *I* think 403 is somewhat (and very technically) wrong.
- 404 is kinda technically correct.
- An empty result `""` is definitely wrong, since it's not even valid JSON.
- An empty object `"{}"` would be correct if everything is optional, which it isn't, and I don't think we want to make it.
- An empty array `"[]"` would be correct if we changed the expected return to be an array, but I really don't think we want to do that.
- `null` would be correct only if we explicitly said that the result could be "a JSON account object or null". Which, I guess would be fine.

So... I think 404 is best, followed by `null`?

Yet another option (because clearly we don't have enough) would be to define an implicit account for the unauthenticated that always has to be there. That would of course be fine, but I think I prefer 404 and `null` over that?

eldering: I think using 404 or 403 is better than returning weird JSON responses (either `null` or `{}`) that don't match the normal specification of this endpoint. If we're to return an exceptional state (no account), I think it makes most sense to do so using an error code.

deboer-tim: Ok, I updated to 404.

nickygerritsen (approved):

Contest_API.md
```diff
@@ -1178,16 +1178,17 @@ Returned data:
 
 ### Accounts
 
-The accounts used for access to the contest.
+The accounts used for accessing the contest, including the account representing the client.
```
niemela, line 1181: A bit of a nitpick:

This wording is strange. Of course a list of all the accounts used for accessing the contest would include the account currently used. 

Also, I think "currently used" is better than "representing the client"?

I *think* what you want to say is something like:
>  The accounts used for accessing the contest, as well as information about the account currently used for accessing the API.

?

Contest_API.md
```diff
@@ -1178,16 +1178,17 @@ Returned data:
...
 
 The following endpoints are associated with accounts:
 
 | Endpoint                       | Mime-type        | Required? | Description
 | :----------------------------- | :--------------- | :-------- | :----------
 | `/contests/<id>/accounts`      | application/json | yes       | JSON array of all accounts with properties as defined in the table below.
 | `/contests/<id>/accounts/<id>` | application/json | yes       | JSON object representing a single account with properties as defined in the table below.
+| `/contests/<id>/account`       | application/json | yes       | JSON object representing a single account representing the client making the request, with properties as defined in the table below.
```
niemela, line 1189: The double "representing" is awkward.

niemela (approved):
I think the changes I suggest are worth doing, but they are not really blockers...

niemela (approved):

Contest_API.md
deboer-tim: I didn't like any of the synonyms I found, so went with "of", i.e. "single account of the client".

Contest_API.md
deboer-tim: Changed.

nickygerritsen (approved):

eldering (approved):

PR #83, opened 2022-03-10
Capabilities and Access

deboer-tim:
This change adds capabilities, and the /access endpoint to view them:
- Added overall description of capabilities along with a table and links.
- Changed Roles section to Authentication and updated it. Used the word 'kind' to avoid having more 'types'. :)
- Expanded extensibility section to be clear that unknown contest types (and notifications of them) and capabilities must be allowed. Moved it after notifications since it refers to that (and felt like it should be the last topic in design principles).
- Updated the contest, submissions, and clarifications endpoints to describe their behaviours as capabilities. An explicit TODO is left in to complete the other capabilities in a following PR. I wouldn't normally leave the API in an incomplete state, but this change is large enough already and I'd prefer to agree on how we document capabilities before rolling it out everywhere.
- Added new /access endpoint for clients to see what capabilities they have access to and visibility of endpoints/contest types and their properties.
- Found a few references to 'user' and changed them to 'client'.
- Tried to remove the word 'role' in most cases, since we now have accounts with capabilities instead of specific roles.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -43,7 +43,7 @@ The interface is implemented as a HTTP REST interface that outputs
 information in [JSON](https://en.wikipedia.org/wiki/JSON) format
 ([RFC 7159](https://tools.ietf.org/html/rfc7159)). This REST interface 
 should be provided over HTTPS to guard against eavesdropping on
-sensitive contest data and authentication credentials (see roles below).
+sensitive contest data and [authentication](#authentication) credentials.
 
 ### Endpoint URLs
 
@@ -164,7 +164,7 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
 When there is a failure using any method the response message body
 must include a JSON object that contains the properties 'code' (a number,
 identical to the HTTP status code returned) and 'message' (a string) with
-further information suitable for the user making the request, as per the
+further information suitable for the client making the request, as per the
 following example:
 
 ```json
@@ -172,25 +172,25 @@ following example:
  "message":"Teams cannot send clarifications to another team"}
  ```
 
-### Roles
+### Authentication
 
-Access to this API is controlled via user roles. The API provider must
-require authentication to access each role except for optionally the
-public role. The API provider must support [HTTP basic
+The API provider may allow unauthenticated access to public information,
+but all other access to the API is controlled via authenticated accounts. 
+The API provider must support [HTTP basic
 authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)
 ([RFC](https://tools.ietf.org/html/rfc7617)). This provides a standard
 and flexible method; besides HTTP basic auth, other forms of
 authentication can be offered as well.
 
-Each provider must support at least the following roles, although
-additional roles may be supported for specific uses:
+Each provider must support at least the following kinds of accounts, although
+additional ones may be supported for specific uses:
 
-  - public (default role: contest data that's available to everyone)
+  - public (default access: contest data that's available to everyone)
   - admin (data or capability only available to contest administrators)
 
-Role-based access may completely hide some objects from the user, may
+Each kind of account may completely hide some objects from the client, may
 omit certain properties, or may embargo or omit objects based on the
-current contest time. By default, the public user has read-only access
+current contest time. By default, the public has read-only access
 (no `POST`, `PUT`, `PATCH` or `DELETE` methods allowed) and does
 not have access to judgements and runs from submissions made after the
 contest freeze time.
@@ -305,22 +305,36 @@ with JSON data
 }
 ```
 
-### Extensibility
+### Capabilities
 
-This specification is meant to cover the basic data of contests, with
-the idea that server/client implementations can extend this with more
-data and/or roles. In particular, this specification already lists some
-endpoints or specific properties as optional. The following guidelines
-are meant to ease extensibility.
+Instead of strict role-based access control, this API defines several
+capabilities that define behaviours that clients can expect and
+actions they can perform. For instance, a team account will typically
+have access to a "team_submit" capability that allows a team to perform
+POST operations on the submissions endpoint, but doesn't allow it to
+set the submission id or timestamp; an administrator may have access
+to a "contest_start" capability that allows it to PATCH the start
+time of the contest. These coarse-grained capabilities allow more
+flexibility for contest administrators and tools to define capabilities
+that match the requirements of a specific contest - e.g. whether teams
+can submit clarifications or not.
 
-  - Clients should accept extra properties in endpoints, that are not
-    specified here.
-  - Servers should not expect clients to recognize more than the basic,
-    required specification.
-  - In this specification and extensions, a property with value `null`
-    may be left out by the server (i.e. not be present). A client must
-    treat a property with value `null` equivalently as that property
-    not being present.
+All capabilities are listed in the table below, and are defined
+inline with each endpoint. Clients can use
+the [Access](#access) endpoint to see which capabilities they have
+access to.
+
+| Capability                                | Description                                  |
+| :---------------------------------------- | :------------------------------------------- |
+| [contest_start](#contest-capabilties)     | Control the contest's start time             |
+| [team_submit](#submission-capabilities)   | Submit as a team                             |
+| [team_clar](#clarification-capabilities)  | Submit clarifications as a team              |
+| [proxy_submit](#submission-capabilities)  | Submit as a shared team proxy                |
+| [proxy_clar](#clarification-capabilities) | Submit clarifications as a shared team proxy |
+| [admin_submit](#submission-capabilities)  | Submit as an admin                           |
+| [admin_clar](#clarification-capabilities) | Submit clarifications as an admin            |
+
+TODO - add capabilities related to team view, awards, and freeze time.
 
 ### Notification format
 
@@ -355,7 +369,7 @@ etc.) or entire collection. The general format for events is:
 
 All event types correspond to an API endpoint, as specified in the table below.
 
-| Event           | API Endpoint                          |
+| Type            | API Endpoint                          |
 | :-------------- | :------------------------------------ |
 | contest         | `/contests/<id>`                      |
 | judgement-types | `/contests/<id>/judgement-types/<id>` |
@@ -458,6 +472,24 @@ Means that endpoint `/contests/<contest_id>/submissions/187` has been updated to
    "files": [{"href":"contests/wf14/submissions/187/files","filename":"files.zip","mime":"application/zip"}]
 }
 ```
+
+### Extensibility
+
+This specification is meant to cover the basic data of contests, with
+the idea that server/client implementations can extend this with more
+types, properties, and/or capabilities. In particular, this specification
+already lists some endpoints or specific properties as optional. The
+following requirements are meant to ease extensibility further:
+
+  - Clients should accept additional (unknown) event types in notifications.
+  - Clients should accept additional (unknown) properties in endpoints.
+  - Clients should accept additional (unknown) capabilities.
+  - Servers should not expect clients to recognize more than the basic,
+    required specification.
+  - In this specification and extensions, a property with value `null`
+    may be left out by the server (i.e. not be present). A client must
+    treat a property with value `null` equivalently as that property
+    not being present.
 
 ## Interface specification
 
@@ -607,19 +639,19 @@ The `countdown_pause_time` may change to indicate approximate delay.
 Countdown is resumed by setting a new `start_time` and resetting
 `countdown_pause_time` to `null`.
 
-#### PATCH start\_time
+#### Contest capabilties
 
-Implementations must have a role that has the ability to clear or set the
-contest start time via a PATCH method.
+Clients with the `contest_start` [capability](#capabilities) have the ability to
+set or clear the contest start time via a PATCH method.
 
 The PATCH must include a valid JSON object with only two or three
-properties allowed: the contest id (used for verification), a
-start\_time (a `<TIME>` value or `null`), and an optional
-countdown\_pause\_time (`<RELTIME>`). As above, countdown\_pause\_time
+properties allowed: the contest `id` (used for verification), a
+`start_time` (a `<TIME>` value or `null`), and an optional
+`countdown_pause_time` (`<RELTIME>`). As above, `countdown_pause_time`
 can only be non-null when start time is null.
 
-The request should fail with a 401 error code if the user does not have sufficient
-access rights, or a 403 error code if the contest is started or within 30s of
+The request should fail with a 401 error code if the client does not have the
+required capability, or a 403 error code if the contest is started or within 30s of
 starting, or if the new start time is in the past or within 30s.
 
 #### Examples
@@ -1201,7 +1233,7 @@ Properties of account objects:
 | people\_id        | ID      | no        | yes       | The person that this account is for, if the account is only for one person.
 
 Accounts exist in the API primarily for configuration from a contest archive, or an administrator comparing one CCS to another. It is
-expected that non-admin roles never see passwords, and typically do not see accounts other than their own.
+expected that non-admin clients never see passwords, and typically do not see accounts other than their own.
 
 The account endpoint exists so that the clients can tell which account (and hence which person or team) they are logged in as. It is not
 expected to exist in a contest archive, and will fail with a 404 error code if the client is not authenticated.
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1337,13 +1440,20 @@ zip archive. These must be stored directly from the root of the zip
 file, i.e. there must not be extra directories (or files) added unless
 these are explicitly part of the submission content.
 
-#### POST and PUT submissions
+#### Submission capabilities
```
nickygerritsen, line 1443: I'm not sure if I like renaming this sections (also above and below)

Contest_API.md
```diff
@@ -164,33 +164,33 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
+but all other access to the API is controlled via authenticated accounts. 
+The API provider must support [HTTP basic
 authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)
 ([RFC](https://tools.ietf.org/html/rfc7617)). This provides a standard
 and flexible method; besides HTTP basic auth, other forms of
 authentication can be offered as well.
 
-Each provider must support at least the following roles, although
-additional roles may be supported for specific uses:
+Each provider must support at least the following kinds of accounts, although
```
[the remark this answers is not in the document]
deboer-tim: I don't really want to try changing that in this PR, but I agree that maybe it doesn't belong here (or maybe at even in the CCS' spec? The way we run finals I don't think we even use the CCS public access).

Contest_API.md
```diff
@@ -607,19 +639,19 @@ The `countdown_pause_time` may change to indicate approximate delay.
...
-start\_time (a `<TIME>` value or `null`), and an optional
-countdown\_pause\_time (`<RELTIME>`). As above, countdown\_pause\_time
+properties allowed: the contest `id` (used for verification), a
+`start_time` (a `<TIME>` value or `null`), and an optional
+`countdown_pause_time` (`<RELTIME>`). As above, `countdown_pause_time`
 can only be non-null when start time is null.
 
-The request should fail with a 401 error code if the user does not have sufficient
-access rights, or a 403 error code if the contest is started or within 30s of
+The request should fail with a 401 error code if the client does not have the
```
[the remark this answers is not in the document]
deboer-tim: Agreed. This isn't changed by this PR, but maybe it should be something like a 405?

Contest_API.md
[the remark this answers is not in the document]
nickygerritsen: Yeah let’s not change it now but discuss at a call

Contest_API.md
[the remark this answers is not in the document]
nickygerritsen: 🤷‍♂️ Let’s also discuss later

Contest_API.md
deboer-tim: I see your point, but I don't love using HTTP methods as section titles and think the document will be easier to scan if we have a standard pattern for each section that talks about any POST/PUT/PATCH/DELETE options. Maybe something like "Modifying submissions" (and e.g. "Modifying awards" in the future), or (since we have "Examples" for each type) leaving the type name out and use something like "Editing"?

nickygerritsen: I like your last two suggestions way more

Contest_API.md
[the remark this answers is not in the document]
niemela: I agree with both of you. I.e. we *should* (probably) do the change @nickygerritsen suggests, but we shouldn't do it in this PR.

Contest_API.md
```diff
@@ -1337,13 +1440,20 @@ zip archive. These must be stored directly from the root of the zip
 file, i.e. there must not be extra directories (or files) added unless
 these are explicitly part of the submission content.
 
-#### POST and PUT submissions
+#### Submission capabilities 
+
+To add a submission clients can use the `POST` method on the submissions endpoint or the
```
eldering, line 1445: Nitpick: add a comma here for readability:
```suggestion
To add a submission, clients can use the `POST` method on the submissions endpoint or the
```

Contest_API.md
```diff
@@ -1337,13 +1440,20 @@ zip archive. These must be stored directly from the root of the zip
...
 
-To add a submission one can use the `POST` method on the submissions endpoint or the
-`PUT` method directly on an object url. `POST` is typically used by teams submitting
-to the contest and `PUT` is used by admin users or tools.
-Both must include a valid JSON object with the same properties the submission
-endpoint returns with a `GET` request with the following exceptions:
+| Name              | Description
+| :---------------- | :----------
+| team_submit       | POST a submissision as a team
+| proxy_submit      | POST a submission as a proxy (able to submit on behalf of a team)
```
eldering, line 1452: For clarity, I'd propose
```suggestion
| proxy_submit      | POST a submission as a proxy (able to submit on behalf of any team)
```

Contest_API.md
```diff
@@ -1337,13 +1440,20 @@ zip archive. These must be stored directly from the root of the zip
...
+| admin_submit      | POST or PUT a submission as an admin
```
eldering, line 1453: maybe add something along "This allows overriding submission id, team and time among other things"? Or refer to the explanation below on how this can be used?

Contest_API.md
niemela: This would precludes a proxy (i.e. an account) that has the right to submit for some, but not all, teams. This would make sense e.g. in the original use case we imagined for proxies, i.e. multi-site contests.

Contest_API.md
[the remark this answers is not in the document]
deboer-tim: Added to CLICS call agenda.

Contest_API.md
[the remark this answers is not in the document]
deboer-tim: Added to CLICS agenda.

Contest_API.md
deboer-tim: Agreed. I hope neither of you want to add API for the proxy to tell which team(s) it can submit for? :)

In practise I think it's likely that we wouldn't have to restrict a proxy to a subset of teams b/c only the teams it can submit for will have either network or authentication access to it, but I agree we should't preclude this in the API. How about "on behalf of team(s)"? That implies it may be more than 1 but doesn't imply all teams.

deboer-tim: I was pushing another commit so I went with "team(s)", let me know if it works.

Contest_API.md
deboer-tim: I just did a commit with a "Modifying x" pattern, partially b/c if I went with a single word I don't know how to link to each section in markdown.  ;-)

Contest_API.md
niemela: > hope neither of you want to add API for the proxy to tell which team(s) it can submit for? :)

Nope :).

Contest_API.md
```diff
@@ -164,33 +164,33 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
 
-  - public (default role: contest data that's available to everyone)
+  - public (default access: contest data that's available to everyone)
   - admin (data or capability only available to contest administrators)
 
-Role-based access may completely hide some objects from the user, may
+Each kind of account may completely hide some objects from the client, may
 omit certain properties, or may embargo or omit objects based on the
-current contest time. By default, the public user has read-only access
+current contest time. By default, the public client has read-only access
```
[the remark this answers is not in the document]
deboer-tim: Rewrote public access and auth section.

Contest_API.md
deboer-tim: Added "descriptions below" above both tables.

Contest_API.md
```diff
@@ -1352,16 +1458,16 @@ endpoint returns with a `GET` request with the following exceptions:
...
   must match the ID of the team associated with the request.
-* Simple proxies using an `admin` role should also use `POST` to let the server
+* The `proxy_submit` capability also uses `POST` and lets the server
   determine the `id`. `team_id` must be provided but `time` must not and will be
   determined by the server.
-* For more advanced scenarios an `admin` role may use a `PUT`. `time` is required and
-  the client is responsible for making sure the provided `id` is unique. For example
-  in a setup with a central CCS with satellite sites
+* For more advanced scenarios the `admin_submit` capability may use a `POST` (must not
+  include an `id`) or `PUT` (client is required to include a unique `id`). In both
```
niemela, line 1469: So if you have `proxy_submit` you may provide `id`in a POST (which feels wrong in itself. If you are deciding the `id` you should use a PUT not a POST), but `admin_submit` may not. This feels inconsistent.

I suggest that neither can include an `id` in their POST, but both are allowed to PUT with an `id`.

Contest_API.md
```diff
@@ -1598,37 +1705,47 @@ Properties of clarification message objects:
...
-  required to either be absent or present depending on the use case, e.g.
-  whether the server is the CCS, is acting as a proxy, or a caching
-  proxy. See notes under the submission interface for more detail. In cases where
-  these properties are not allowed the server will respond with a 4xx error code.
+* The `proxy_clar` capability also uses `POST` and lets the server
+  determine the `id`. `team_id` must be provided but `to_team_id`, `time`, and
+  `contest_time` must not and will be determined by the server.
+* The `admin_clar` capability may use a `POST` (must not
+  include an `id`) or `PUT` (client is required to include a unique `id`).
+  In both cases `time` is required.
```
niemela, line 1735: Same comment as for submissions.

niemela (changes requested):

niemela (approved):

Contest_API.md
```diff
@@ -41,9 +41,7 @@ document are to be interpreted as described in
 
 The interface is implemented as a HTTP REST interface that outputs
 information in [JSON](https://en.wikipedia.org/wiki/JSON) format
-([RFC 7159](https://tools.ietf.org/html/rfc7159)). This REST interface 
-should be provided over HTTPS to guard against eavesdropping on
-sensitive contest data and authentication credentials (see roles below).
+([RFC 7159](https://tools.ietf.org/html/rfc7159)).
```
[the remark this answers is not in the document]
deboer-tim: Technically, we only said "should" for HTTPS before, and it only stands out b/c I made it a must for non-public data. :)

I've backed this out with a slight rewording, so it's still a should in this section and then a must for anything sensitive or requiring auth lower down. This ends up with a slight text duplication between these sections. If you prefer we could either make HTTPS a must for anything, or focus on all the cases that need HTTPS (auth, non-sensitive data, etc) and not mention that you could do HTTP for anything not on the list.

(And yeah, I know your feelings on REST ;-) )

Contest_API.md
```diff
@@ -164,36 +162,34 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
  "message":"Teams cannot send clarifications to another team"}
  ```
 
-### Roles
+### HTTPS and Authentication
 
-Access to this API is controlled via user roles. The API provider must
-require authentication to access each role except for optionally the
-public role. The API provider must support [HTTP basic
+The API provider may allow HTTP or unauthenticated access for information
```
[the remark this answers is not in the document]
deboer-tim: HTTP is fine for public data, HTTPS just hides what you're looking at from snooping and makes it harder to sit in the middle. Just polled some common sites though, and doing everything over HTTPS does seem more common now, so as per previous comment we could make that requirement now.

Contest_API.md
```diff
@@ -164,36 +162,34 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
-
-Role-based access may completely hide some objects from the user, may
-omit certain properties, or may embargo or omit objects based on the
-current contest time. By default, the public user has read-only access
-(no `POST`, `PUT`, `PATCH` or `DELETE` methods allowed) and does
-not have access to judgements and runs from submissions made after the
-contest freeze time.
+Depending on the client account's access, some objects may be completely
+hidden from the client, may omit certain properties, or may embargo or
+omit objects based on the current contest time.
```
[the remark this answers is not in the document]

Contest_API.md
```diff
@@ -305,22 +301,36 @@ with JSON data
...
-    specified here.
-  - Servers should not expect clients to recognize more than the basic,
-    required specification.
-  - In this specification and extensions, a property with value `null`
-    may be left out by the server (i.e. not be present). A client must
-    treat a property with value `null` equivalently as that property
-    not being present.
+### Capabilities
+
+Instead of strict role-based access control, this API defines several
```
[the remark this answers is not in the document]

Contest_API.md
```diff
@@ -355,7 +365,7 @@ etc.) or entire collection. The general format for events is:
 
 All event types correspond to an API endpoint, as specified in the table below.
 
-| Event           | API Endpoint                          |
+| Type            | API Endpoint                          |
```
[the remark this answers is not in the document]
deboer-tim: Yeah, I know. It is 'event type' and in the original context 'event' is better, but when referred to from /access I thought 'type' was a better fit to cover both. I didn't really want to move the table to somewhere common, can't think of any better option.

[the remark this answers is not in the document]
niemela: Yes, that's also a good point.

I can't think of a better option either.

Contest_API.md
```diff
@@ -164,36 +164,31 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
-
-Role-based access may completely hide some objects from the user, may
-omit certain properties, or may embargo or omit objects based on the
-current contest time. By default, the public user has read-only access
-(no `POST`, `PUT`, `PATCH` or `DELETE` methods allowed) and does
-not have access to judgements and runs from submissions made after the
-contest freeze time.
+Depending on the client account's access, some objects may be completely
+hidden from the client, may omit certain properties, or may embargo or
+omit objects based on the current state of the contest.
```
eldering, line 191: Small rewrite to active voice and with consistent subject/object:
```suggestion
Depending on the client account's access, the API provider may completely
hide some objects from the client, may omit certain properties, or may embargo or
omit objects based on the current state of the contest.
```

Contest_API.md
```diff
@@ -305,22 +300,36 @@ with JSON data
...
-  - Servers should not expect clients to recognize more than the basic,
-    required specification.
-  - In this specification and extensions, a property with value `null`
-    may be left out by the server (i.e. not be present). A client must
-    treat a property with value `null` equivalently as that property
-    not being present.
+### Capabilities
+
+The API specifies several
+capabilities that define behaviours that clients can expect and
```
eldering, line 306: Use American English
```suggestion
capabilities that define behaviors that clients can expect and
```

Contest_API.md
```diff
@@ -305,22 +300,36 @@ with JSON data
...
+actions they can perform. For instance, a team account will typically
+have access to a "team_submit" capability that allows a team to perform
+POST operations on the submissions endpoint, but doesn't allow it to
+set the submission id or timestamp; an administrator may have access
+to a "contest_start" capability that allows it to PATCH the start
+time of the contest. These coarse-grained capabilities allow more
+flexibility for contest administrators and tools to define capabilities
+that match the requirements of a specific contest - e.g. whether teams
```
eldering, line 314: This is really nitpicky, but I think a comma is better here, or at least more common. Also if used, should this be an em-dash (or an en-dash)?
```suggestion
that match the requirements of a specific contest, e.g. whether teams
```

Contest_API.md
```diff
@@ -1240,6 +1266,77 @@ Returned data:
...
+client in a live contest, and hence will not exist in a contest archive.
+
+This information is provided so that clients know what endpoints are available,
+what notifications may happen, and what capabilities they have, regardless
+of whether objects currently exist or the capability is currently active.
+For instance, a client logged in with a team account would see the problems type and
+team_submit capability before a contest starts, even through they cannot
+see any problems nor submit yet.
+Clients are not expected to call this endpoint more than once
+since the response would not normally change during a contest.
```
eldering, line 1304: Maybe better to change "would" to "should" or even "must"?

Contest_API.md
```diff
@@ -1381,7 +1485,7 @@ The request must fail with a 4xx error code if any of the following happens:
 The response will contain a `Location` header pointing to the newly created submission
 and the response body will contain the initial state of the submission.
 
-Performing a `POST` or `PUT` by any roles other than `admin` and `team` is not supported.
+Performing a `POST` or `PUT` is not supported except when the capabilities are available.
```
eldering, line 1488: Do we actually need to state this here, as it's already said at the top of this section?
If so, I think it would be better to write
```suggestion
Performing a `POST` or `PUT` is not supported when these capabilities are not available.
```
to remove the ambiguity that you might need more or all of these capabilities.

Contest_API.md
```diff
@@ -1598,37 +1703,47 @@ Properties of clarification message objects:
...
+| :---------------- | :----------
+| team_clar         | POST a clarification as a team
+| proxy_clar        | POST a clarification as a proxy (able to submit on behalf of team(s))
+| admin_clar        | POST or PUT a clarification as an admin
+
+All requests must include a valid JSON object with the same properties as the clarifications
+endpoint returns from a `GET` request with the following exceptions:
 
 * When a property value would be null it is optional - you do not need to include it.
   e.g. if a clarification is not related to a problem you can chose to include or
```
eldering, line 1722: Unrelated typo fix:
```suggestion
  e.g. if a clarification is not related to a problem you can choose to include or
```

Contest_API.md
```diff
@@ -1598,37 +1703,47 @@ Properties of clarification message objects:
...
 * The supplied problem, from_team, to_team, or reply_to cannot be found or are not
-  visible to the role that's submitting.
+  visible to the client that's submitting.
 * The provided `id` already exists or is otherwise not acceptable.
 
 The response will contain a `Location` header pointing to the newly created clarification
 and the response body will contain the initial state of the clarification.
 
-Performing a `POST` or `PUT` by any roles other than `admin` and `team` is not supported.
+Performing a `POST` or `PUT` is not supported except when the capabilities are available.
```
eldering, line 1746: Same comment as for submit: do we need to state this again?

Contest_API.md
```diff
@@ -164,36 +164,31 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
  "message":"Teams cannot send clarifications to another team"}
  ```
 
-### Roles
+### Authentication
 
-Access to this API is controlled via user roles. The API provider must
-require authentication to access each role except for optionally the
-public role. The API provider must support [HTTP basic
+The API provider may allow unauthenticated access for information
```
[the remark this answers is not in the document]
niemela: I think either works?

Contest_API.md
```diff
@@ -164,36 +164,31 @@ fail or implement a cascading delete (to maintain [referential integrity](#refer
...
+that is fully public - i.e. may be visible to everyone including
```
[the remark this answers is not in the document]
niemela: I think either *works*, but comma is probably better.

Contest_API.md
niemela: I actually don't think the active voice is better here. It's not that the server after careful consideration decides to hide some objects, but rather that given the capabilities (I almost said role) of the client and the state of the contest, some objects are (and should be) hidden.

Contest_API.md
niemela: Definitely not "must".

Contest_API.md
niemela: The rewrite is clearer. :+1:

I do think it makes sense to state this here though.

niemela (dismissed):

Contest_API.md
nickygerritsen: Agreed. If you somehow make a mistake with setting up capabilities for a user you should be able to fix that mistake without being illegal.

nickygerritsen (dismissed):

Contest_API.md
eldering: FYI, my reason for using the active voice was mostly to make the sentence shorter, and I think, slightly easier to read. It's not meant to change any intented meaning/intonation. Also the original sentence was still written half/half in active/passive voice. I'm also fine with completely writing it in passive voice.

deboer-tim: I'm going to commit just b/c it makes the wording consistent. If anyone disagrees we can change the whole thing to passive voice.

Contest_API.md
deboer-tim: Ditto, not 'must'. If you felt strongly I'd be ok with 'should', but otherwise I think this is ok.

nickygerritsen (approved):

Contest_API.md
eldering: Ok, I agree that "must" is too strong, but I think that changing it to "should" would be better, so that clients also have a soft guarantee that they don't need to query this endpoint more than once.

deboer-tim: Ok, changed.

nickygerritsen (approved):

Contest_API.md
```diff
@@ -186,8 +186,8 @@ authentication](https://en.wikipedia.org/wiki/Basic_access_authentication)
 and flexible method; besides HTTP basic auth, other forms of
 authentication can be offered as well.
 
-Depending on the client account's access, some objects may be completely
-hidden from the client, may omit certain properties, or may embargo or
+Depending on the client account's access, the API provider may completely
+hide some objects from the client, may omit certain properties, or may embargo or
```
niemela, line 190: I actually don't like this change...

niemela (approved):

PR #89, opened 2022-03-23
Inherently nullable properties

deboer-tim:
Changing all "inherently nullable" properties to actually allow null. These are all cases where we had a 'double no' for required and nullable before, or the property is only required in some cases. In all of them the client would already need to accept missing or null values (which we say are the same) or the server might have to do some odd things to support (e.g. have /access check if a contest has a formal_name to decide whether or not to include that property). Allowing this is easier for both parties.

Also fixed formatting of contest state table.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -695,7 +695,7 @@ Properties of contest objects:
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
 | scoreboard\_type             | string ?        | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
+| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
 | banner                       | array of FILE ? | Banner for this contest, intended to be an image with a large aspect ratio around 8:1. Only allowed mime types are image/*.
 | logo                         | array of FILE ? | Logo for this contest, intended to be an image with aspect ratio near 1:1. Only allowed mime types are image/*.
 | location.latitude            | number ?        | Latitude in degrees. Required iff location.longitude is present.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -689,13 +689,13 @@ Properties of contest objects:
 | :--------------------------- | :-------------- | :----------
 | id                           | ID              | Identifier of the current contest.
 | name                         | string          | Short display name of the contest.
-| formal\_name                 | string          | Full name of the contest.
+| formal\_name                 | string ?        | Full name of the contest.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
 | scoreboard\_type             | string ?        | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
+| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
 | banner                       | array of FILE ? | Banner for this contest, intended to be an image with a large aspect ratio around 8:1. Only allowed mime types are image/*.
 | logo                         | array of FILE ? | Logo for this contest, intended to be an image with aspect ratio near 1:1. Only allowed mime types are image/*.
 | location.latitude            | number ?        | Latitude in degrees. Required iff location.longitude is present.
@@ -1355,7 +1355,7 @@ The following endpoints are associated with state:
 Properties of state objects:
 
 | Name             | Type   | Description
-| :--------------- | :----- | :-------- | :-------- | :----------
+| :--------------- | :----- | :----------
 | started          | TIME ? | Time when the contest actually started, or `null` if the contest has not started yet. When set, this time must be equal to the [contest](#contests) `start_time`.
 | frozen           | TIME ? | Time when the scoreboard was frozen, or `null` if the scoreboard has not been frozen. Required iff `scoreboard_freeze_duration` is present in the [contest](#contests) endpoint.
 | ended            | TIME ? | Time when the contest ended, or `null` if the contest has not ended. Must not be set if started is `null`.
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -689,13 +689,13 @@ Properties of contest objects:
 | :--------------------------- | :-------------- | :----------
 | id                           | ID              | Identifier of the current contest.
 | name                         | string          | Short display name of the contest.
-| formal\_name                 | string          | Full name of the contest.
+| formal\_name                 | string ?        | Full name of the contest.
```
niemela, line 692: Should have a default.

```suggestion
| formal\_name                 | string ?        | Full name of the contest. Defaults to value of `name`.
```

nickygerritsen (dismissed):

Contest_API.md
```diff
@@ -1613,7 +1613,7 @@ Properties of judgement objects:
 | id                   | ID        | Identifier of the judgement.
 | submission\_id       | ID        | Identifier of the [submission](#submissions) judged.
 | judgement\_type\_id  | ID ?      | The [verdict](#judgement-types) of this judgement. Required iff judgement has completed.
-| score                | number    | Score for this judgement. Required iff contest:scoreboard_type is `score`.
+| score                | number ?  | Score for this judgement. Required iff contest:scoreboard_type is `score`.
```
niemela, line 1616: I don't think this should be nullable. If you don't want to provide a score you must say so in `/access`.

nickygerritsen: Wait but that means that you can not really change the scoring type of a contest, right? If it is `score` you want to give a score (clearly) but if you then change it, you need to update `/access`. Which is 'allowed' but not really preferred.

niemela: Yes, but changing `scoring_type` is also "allowed" and *definitely not referred. So, that sounds about right to me.

nickygerritsen: OK fair.

But I don't really like that the access depends on the contest type, but that might just be me.

Does this for example also mean that `team.group_ids` should be required (it has a `?` now) since you can do that with access? And some fields on the scoreboard related to pass-fail/score types.

niemela: > But I don't really like that the access depends on the contest type, but that might just be me.

Are you thinking that `/access` could be a constant for some/most/a system and then irrelevant properties would just be nulled? That would certainly be easier for the system, but it would also make `/access` significantly less useful.

(For the same reason, I'm not so sire `formal_name` actually should be nullable either. Isn't it better to drop it from `/access` if you don't want to provide it? That said, the nullable-or-not question is a lot less important for the contest endpoint since it would have to already be present for `/access` to exist anyway.)

Maybe we need a system level (as opposed to contest level) access endpoint as well. That will give you a superset of what could be offered at a contest on the system. I.e. `/access` would tell you the maximum set of "stuff" that a give API provider could ever offer and `/contests/<id>/access` would tell you what any particular contest is *actually* going to offer. I'm only 50% joking.

nickygerritsen: Joking aside we should also change group_ids and the scoreboard properties to not be billable if we go this way

niemela: Yes, agreed.

deboer-tim: If a property isn't in /access then you're guaranteed not to see one, but I was assuming that we wouldn't guarantee the reverse. e.g. If org.logo is not returned from /access then you are assured that the provider doesn't support logos or there are none in this contest. If org.logo is returned then the provider definitely supports them - but may not have scanned every org in the current contest to see if there are any. Or maybe there aren't any currently but it doesn't want to commit that there won't be any later, or hasn't done anything to block an inadvertent change.

Said another way, it gets more and more complex to handle account-specific and then contest-specific /access, and I personally don't expect all of our current API providers to get this far and get it right. In these cases there isn't even any benefit from implementing it - e.g. clients should check if this is a scoring contest by looking at contest.scoreboard_type not whether judgements.score exists, and in all of the rest the client would need to have null checks anyway.

Contest_API.md
```diff
@@ -689,13 +689,13 @@ Properties of contest objects:
...
-| formal\_name                 | string          | Full name of the contest.
+| formal\_name                 | string ?        | Full name of the contest.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
 | scoreboard\_type             | string ?        | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
+| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
```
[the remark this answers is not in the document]
deboer-tim: I pushed the same change as #90, so we don't need both now.

Contest_API.md
deboer-tim: I hesitated with this because what we said before is that clients should fall back to using name (i.e. 'the contest doesn't have a formal name and you should use the name for display purposes'), which isn't quite the same as defaulting formal_name to name. To me this risks someone writing 'if (formal_name == null) then formal_name = name' or saying 'the contest has both properties and they're identical'.

niemela: To me "clients should fall back to using name" means the same as "defaulting formal_name to name". 

If we don't say that formal_name defaults to name, then if formal_name is missing that means that there is no formal_name, and it must be up to the client to choose how to handle that, we shouldn't tell the client what it "should" do.

Contest_API.md
niemela: > If a property isn't in /access then you're guaranteed not to see one, but I was assuming that we wouldn't guarantee the reverse

Right, but only because if some of the properties listed in `/access` are nullable then they could be null (duh) and since we say that null and missing must be equal, they must be able to be missing.

Contest_API.md
deboer-tim: Would "If formal_name does not exist use name for display purposes" be enough? I'm fine telling the client what to do in this case, I just don't like (possibly due to history?) formal_name picking up the value of name.

eldering: I think given the discussion we had, this should not be nullable: either the server has the formal name and then should specify that in `/access` and it should be non-null, or it doesn't, in which case it shouldn't list it in `/access` and then it must not be present.

Contest_API.md
```diff
@@ -689,13 +689,13 @@ Properties of contest objects:
...
+| formal\_name                 | string ?        | Full name of the contest.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
 | scoreboard\_type             | string ?        | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
+| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
```
eldering, line 698: I think this should not be nullable for the same reasons as `formal_name` above.

Contest_API.md
eldering: I think this one should also not be nullable: if you say in `/access` that you'll provide it, then you must always.

deboer-tim:
Invalid/will be replaced as discussed today.

PR #91, opened 2022-03-26
Add logos to API

niemela:


Contest_API.md
```diff
@@ -573,10 +573,12 @@ The following endpoint is associated with API information:
...
 
-| Name        | Type   | Description
-| :---------- | :----- | :----------
-| version     | string | Version of the API. For this version must be the string `2022-07-draft`. Will be of the form `<yyyy>-<mm>`, `<yyyy>-<mm>-draft`, or simply `draft`.
-| version_url | string | Link to documentation for this version of the API.
+| Name        | Type            | Description
+| :---------- | :-------------- | :----------
+| version     | string          | Version of the API. For this version must be the string `2022-07-draft`. Will be of the form `<yyyy>-<mm>`, `<yyyy>-<mm>-draft`, or simply `draft`.
+| version_url | string          | Link to documentation for this version of the API.
+| version_url | string          | Link to documentation for this version of the API.
```
nickygerritsen, line 580: Double version url line

Contest_API.md
```diff
@@ -573,10 +573,12 @@ The following endpoint is associated with API information:
...
+| logo        | array of FILE ? | Logo for this data provider, intended to be an image with aspect ratio near 1:1. Only allowed mime types are image/*. The different files in the array should be different file formats and/or sizes of the same image. If an API combines data from many providers that have logos and wants to represent this it should combine the images into a single image that could then be offered in multiple file formats and/or sizes.
```
nickygerritsen, line 581: I’m not sure if I’m a fan of this combining. How would this work if we have a primary, shadow and CDS? There is currently no real link between primary and shadow on the CDS.

It would also mean a “combiner” needs to do stuff that might be really image dependent. For example combining an svg and png. Or strange aspect ratios. Or background colors.

Contest_API.md
niemela: Doh, thanks. Fixed.

Contest_API.md
niemela: > I’m not sure if I’m a fan of this combining.

Me neither. But everything else is worse.

> How would this work if we have a primary, shadow, and CDS? There is currently no real link between primary and shadow on the CDS.

That's no problem. Primary and shadow would just return their own logos. I would argue that they are both primary data sources. The CDS would/could/should create a combined image of all three.

> It would also mean a “combiner” needs to do stuff that might be really image dependent. For example combining an svg and png. Or strange aspect ratios. Or background colors.

Indeed. OTOH, it's not really at all required. It simply says that you can't represent multiple sources by having different images in the array represent different things. All the files must represent the same images, just in different formats and sizes. You can still choose to combine the images or ignore them.

All that said. The CDN should totally do it. Your move @deboer-tim :).

eldering: I think we don't need to specify the last sentence, and just leave it up to the server (e.g. the CDS) to decide how to handle this: provide its own logo, a merged image, or nothing at all. I do agree with the first part that all images in the array should represent the same visual image.

eldering: I think besides a logo, it would also be nice/useful to add something like `system_name`, that could be used to identify the system. Determining which CCS or CDS you are talking to might be useful to understand what additional features it might support, and it would also be nice for allowing the client to show a string, not just a logo. But this could also be a separate PR.

deboer-tim: I had a similar PR ready to go, and I was going to suggest provider (string) and logo (image), i.e. the name and logo of who is providing the API. I think the string would be useful, e.g. for a cli to say 'Connected to Kattis' or 'You've submitted to DOMjudge'.

I agree we should not include the comment about different images or merging. The first opens up questions about what I could do with other image properties (they don't have to be the same image?) and the second is an implementation choice.

In reality though, I don't think merging images is practical. You'd always end up trying to make a large png. How big do you make the gaps? Deal with slightly different aspect ratios? Make it look ok? Smaller versions would look poor. Need to block API connects on startup until you can connect to every CCS, etc. In the end, this merged image of CDS and 2 CCSs still isn't going to look great compared with the manual presentation we currently use that just shows primary and secondary CCS names and logos in widescreen.

The only solution I've come up with so far is to add a second name and logo to each contest, e.g. a 'host' or 'ccs' name and logo. For a CCS these would return the same as the / endpoint, but for the CDS you'd be able to tell that the CDS is providing the API but still see which CCS is hosting each contest. Maybe this is just an optional property that the CDS adds based on downstream provider.

FWIW, we still don't have any way of telling which CCS is primary and which is secondary - to the CDS it's currently just two different providers with independent contests. I think we just solve this with convention though, e.g. shadow contest id is configured as '<id>-shadow' on the CDS.

niemela: > I think we don't need to specify the last sentence, and just leave it up to the server (e.g. the CDS) to decide how to handle this: provide its own logo, a merged image, or nothing at all.

It *is* currently left up to the server. The last sentence does not actually add *any* additional requirements, it only clarifies a logical implication of what has already been said, and since that implication is arguable a bit subtle, I think that's a very good thing.

It does not *at all* say that you must combine images, just that if you want to provide multiple different images (i.e. different not only in size or format) then you would have to combine them. Do you disagree that this is true and already implied without that last sentence? Do you think it's so obvious that it's not needed?

Would it be better it we turned it around and changed the last sentence to:
> This means that if an API combines data from many providers that have different logos and it wants to represent this, it may not return all of them in the array since they would not be different file formats and/or sizes of the same image.

?

niemela: > I think besides a logo, it would also be nice/useful to add something like `system_name`, that could be used to identify the system.

Yes, that makes sense.

I would argue that it should be `logo` and `name` or `system_logo` and `system_name` (or some other synonym for "system", such as "provider"). I would prefer the first.

I added this.

niemela: > I agree we should not include the comment about different images or merging. 

Not even the "the files must represent the same image" part?

> The first opens up questions about what I could do with other image properties (they don't have to be the same image?) 

Yes, they *definitely* do. Do we need to clarify that everywhere?

> and the second is an implementation choice.

True, but (as stated above) it's also not really an additional requirement.
 
> In reality though, I don't think merging images is practical. 

That might very well be true :(.

> You'd always end up trying to make a large png. How big do you make the gaps? Deal with slightly different aspect ratios? Make it look ok? Smaller versions would look poor. Need to block API connects on startup until you can connect to every CCS, etc. In the end, this merged image of CDS and 2 CCSs still isn't going to look great compared with the manual presentation we currently use that just shows primary and secondary CCS names and logos in widescreen.

How and why is that? The presentation we now *also* has to deal with all those things, and would *also* look bad if aspects ratios are too much off. Also, since the presentation is shown on a screen it *could* obviously be saved as an image...

All that said, combining into an image *does* remove all the flexibility for fitting into different use cases...

> The only solution I've come up with so far is to add a second name and logo to each contest, e.g. a 'host' or 'ccs' name and logo. For a CCS these would return the same as the / endpoint, but for the CDS you'd be able to tell that the CDS is providing the API but still see which CCS is hosting each contest. Maybe this is just an optional property that the CDS adds based on downstream provider.

I would rather add a more complicated structure to the api endpoint in that case. Some thing like a "subprovider" array that contains objects with name and logo. Adding the separate logos to contest would e.g. not support the case if you have a system that provides additional data, but not a *separate* contest. E.g. a analytics add-on.

nickygerritsen: > How and why is that? The presentation we now also has to deal with all those things, and would also look bad if aspects ratios are too much off. Also, since the presentation is shown on a screen it could obviously be saved as an image...

It could. But @deboer-tim handpicked images (for example the one he uses for DOMjudge has a different background) and hardcoded DOMjudge and Kattis. So although in theory it could look bad in practice it doesn't since it's hardcoded.

> I would rather add a more complicated structure to the api endpoint in that case. Some thing like a "subprovider" array that contains objects with name and logo. Adding the separate logos to contest would e.g. not support the case if you have a system that provides additional data, but not a separate contest. E.g. a analytics add-on.

I don't like the subprovider name, but I do like this idea.

niemela: > It could. But @deboer-tim handpicked images (for example the one he uses for DOMjudge has a different background) and hardcoded DOMjudge and Kattis. So although in theory it could look bad in practice it doesn't since it's hardcoded.

Ok, fair, and good, point.

> I don't like the subprovider name, but I do like this idea.

Yes, I also don't like that name.

deboer-tim: > > It could. But @deboer-tim handpicked images (for example the one he uses for DOMjudge has a different background) and hardcoded DOMjudge and Kattis. So although in theory it could look bad in practice it doesn't since it's hardcoded.
> 
> Ok, fair, and good, point.

Ditto, good point. I guess what I was trying to say it that there's at least a higher likelihood of any automated solution looking good if the UI-provider has the data and can decide how to show it (a list, logos with text below or beside, animation, picking fonts, etc) vs an intermediate like the CDS trying to merge.

> > I don't like the subprovider name, but I do like this idea.
> 
> Yes, I also don't like that name.

Ditto, I like the idea of this being flexible.

Still not totally comfortable with this being at the API level though. I know it's not common usage, but if there's an archive of contests (or as we do day 1 of finals and playback a previous contest while we configure the correct one) then there could be 3 CCSs, analytics, and CDS, and you can't tell which is for which contest. Not sure if this is something we should correctly model in the API or just a use-case we shouldn't bother supporting. Also means the CDS can't respond to / until we've connected to (or timed out on) all configured CCSs.

niemela: > Also means the CDS can't respond to / until we've connected to (or timed out on) all configured CCSs.

No, it just means that the CDS needs to update / as it connects to more sources.

deboer-tim: > > I agree we should not include the comment about different images or merging.
> 
> Not even the "the files must represent the same image" part?
> 
> > The first opens up questions about what I could do with other image properties (they don't have to be the same image?)
> 
> Yes, they _definitely_ do. Do we need to clarify that everywhere?
> 
> > and the second is an implementation choice.
> 
> True, but (as stated above) it's also not really an additional requirement.

Mostly I don't like putting these just on this one property, because it implies the same isn't true for other image properties. I feel like the no-different-images requirement would be obvious enough (?) but if we feel the need to say it I'd just move it 'up' to somewhere general.

nickygerritsen: > > Also means the CDS can't respond to / until we've connected to (or timed out on) all configured CCSs.
> 
> No, it just means that the CDS needs to update / as it connects to more sources.

But then we need an event for that? Or clients won't know.

deboer-tim: > > > Also means the CDS can't respond to / until we've connected to (or timed out on) all configured CCSs.
> > 
> > 
> > No, it just means that the CDS needs to update / as it connects to more sources.
> 
> But then we need an event for that? Or clients won't know.

Please nooooo! :)

This reminded me of another point though: this PR puts the response to this under /api in the CAF. That makes total sense for the ability to look at a CAF and tell what version of the spec it matches, but once you start proxying or merging sources like the CDS you'll end getting merged data here too. e.g. CDS configured to expose a contest from Kattis and another from DOMjudge, any client CAF would end up with both logos in both contests, which is a little odd. :-/

nickygerritsen (approved):

eldering (approved):

PR #95, opened 2022-03-28
Mark some properties as non-nullable.

eldering:
This is more a discussion starter than a ready PR, but I think that
these are all properties that should always be present, if the
API provider listed them in the `/access` endpoint, and thus
promised to provide these. These properties cannot really be
missing for reasons out of control of the server.

I'll add more comments inline.

See also #89

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1210,7 +1210,7 @@ Properties of team objects:
 | display\_name     | string ?        | Display name of the team. If not set, a client should revert to using the name instead.
 | organization\_id  | ID ?            | Identifier of the [ organization](#organizations) (e.g. university or other entity) that this team is affiliated to.
 | group\_ids        | array of ID     | Identifiers of the [ group(s)](#groups) this team is part of (at ICPC WFs these are the super-regions). No meaning must be implied or inferred from the order of IDs. The array may be empty. Required iff groups endpoint is available.
-| hidden            | boolean         | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
+| hidden            | boolean         | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to `false`.
 | location.x        | number ?        | Team's x position in meters. Required iff location.y or location.rotation is present.
 | location.y        | number ?        | Team's y position in meters. Required iff location.x or location.rotation is present.
 | location.rotation | number ?        | Team's rotation in degrees. Required iff location.x or location.y is present.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -689,7 +689,7 @@ Properties of contest objects:
 | :--------------------------- | :-------------- | :----------
 | id                           | ID              | Identifier of the current contest.
 | name                         | string          | Short display name of the contest.
-| formal\_name                 | string          | Full name of the contest.
+| formal\_name                 | string ?        | Full name of the contest. Defaults to value of `name`.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -694,8 +694,8 @@ Properties of contest objects:
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
-| scoreboard\_type             | string          | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
+| scoreboard\_type             | string          | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`.
+| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`.
 | banner                       | array of FILE ? | Banner for this contest, intended to be an image with a large aspect ratio around 8:1. Only allowed mime types are image/*.
 | logo                         | array of FILE ? | Logo for this contest, intended to be an image with aspect ratio near 1:1. Only allowed mime types are image/*.
 | location.latitude            | number ?        | Latitude in degrees. Required iff location.longitude is present.
@@ -916,7 +916,7 @@ Properties of language objects:
 | :------------------- | :--------------- | :----------
 | id                   | ID               | Identifier of the language from table below.
 | name                 | string           | Name of the language (might not match table below, e.g. if localized).
-| entry_point_required | boolean          | Whether the language requires an entry point. Defaults to `false`.
+| entry_point_required | boolean          | Whether the language requires an entry point.
 | entry_point_name     | string ?         | The name of the type of entry point, such as "Main class" or "Main file"). Required iff entry_point_required is `true`.
 | extensions           | array of string  | File extensions for the language.
 | compiler             | Command object   | Command used for compiling submissions.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -694,8 +694,8 @@ Properties of contest objects:
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
-| scoreboard\_type             | string ?        | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
+| scoreboard\_type             | string          | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
+| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
 | banner                       | array of FILE ? | Banner for this contest, intended to be an image with a large aspect ratio around 8:1. Only allowed mime types are image/*.
 | logo                         | array of FILE ? | Logo for this contest, intended to be an image with aspect ratio near 1:1. Only allowed mime types are image/*.
 | location.latitude            | number ?        | Latitude in degrees. Required iff location.longitude is present.
@@ -811,7 +811,7 @@ Properties of judgement type objects:
 | :------ | :-------- | :----------
 | id      | ID        | Identifier of the judgement type, a 2-3 letter capitalized shorthand, see table below.
 | name    | string    | Name of the judgement. (might not match table below, e.g. if localized).
-| penalty | boolean ? | Whether this judgement causes penalty time. Required iff contest:penalty\_time is present.
+| penalty | boolean   | Whether this judgement causes penalty time. Required iff contest:penalty\_time is present.
 | solved  | boolean   | Whether this judgement is considered correct.
 
 #### Known judgement types
@@ -916,11 +916,11 @@ Properties of language objects:
 | :------------------- | :--------------- | :----------
 | id                   | ID               | Identifier of the language from table below.
 | name                 | string           | Name of the language (might not match table below, e.g. if localized).
-| entry_point_required | boolean ?        | Whether the language requires an entry point. Defaults to `false`.
+| entry_point_required | boolean          | Whether the language requires an entry point. Defaults to `false`.
 | entry_point_name     | string ?         | The name of the type of entry point, such as "Main class" or "Main file"). Required iff entry_point_required is `true`.
 | extensions           | array of string  | File extensions for the language.
-| compiler             | Command object ? | Command used for compiling submissions.
-| runner               | Command object ? | Command used for running submissions. Relevant e.g. for interpreted languages and languages running on a VM.
+| compiler             | Command object   | Command used for compiling submissions.
+| runner               | Command object   | Command used for running submissions. Relevant e.g. for interpreted languages and languages running on a VM.
 
 Properties of Command objects:
 
@@ -1032,9 +1032,9 @@ Properties of problem objects:
 | ordinal           | number    | A unique number that determines the order the problems, e.g. on the scoreboard.
 | rgb               | string ?  | Hexadecimal RGB value of problem color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet), e.g. `#AC00FF` or `#fff`.
 | color             | string ?  | Human readable color description associated to the RGB value.
-| time\_limit       | number ?  | Time limit in seconds per test data set (i.e. per single run). Should be an integer multiple of `0.001`.
-| test\_data\_count | integer ? | Number of test data sets.
-| max_score         | number ?  | Maximum expected score, although teams may score higher in some cases. Typically used to indicate scoreboard cell color in scoring contests. Required iff contest:scoreboard_type is `score`.
+| time\_limit       | number    | Time limit in seconds per test data set (i.e. per single run). Should be an integer multiple of `0.001`.
+| test\_data\_count | integer   | Number of test data sets.
+| max_score         | number    | Maximum expected score, although teams may score higher in some cases. Typically used to indicate scoreboard cell color in scoring contests. Required iff contest:scoreboard_type is `score`.
 | package           | array of FILE ? | [Problem package](https://www.kattis.com/problem-package-format/). Expected mime type is application/zip. Only exactly one archive is allowed. Not expected to actually contain href for package during the contest, but used for configuration and archiving.
 | statement         | array of FILE ? | Problem statement. Expected mime type is application/pdf. 
 
@@ -1209,8 +1209,8 @@ Properties of team objects:
 | name              | string          | Name of the team.
 | display\_name     | string ?        | Display name of the team. If not set, a client should revert to using the name instead.
 | organization\_id  | ID ?            | Identifier of the [ organization](#organizations) (e.g. university or other entity) that this team is affiliated to.
-| group\_ids        | array of ID ?   | Identifiers of the [ group(s)](#groups) this team is part of (at ICPC WFs these are the super-regions). No meaning must be implied or inferred from the order of IDs. The array may be empty. Required iff groups endpoint is available.
-| hidden            | boolean ?       | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
+| group\_ids        | array of ID     | Identifiers of the [ group(s)](#groups) this team is part of (at ICPC WFs these are the super-regions). No meaning must be implied or inferred from the order of IDs. The array may be empty. Required iff groups endpoint is available.
+| hidden            | boolean         | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
 | location.x        | number ?        | Team's x position in meters. Required iff location.y or location.rotation is present.
 | location.y        | number ?        | Team's y position in meters. Required iff location.x or location.rotation is present.
 | location.rotation | number ?        | Team's rotation in degrees. Required iff location.x or location.y is present.
@@ -1296,7 +1296,7 @@ Properties of account objects:
 | id                | ID        | Identifier of the account.
 | username          | string    | The account username.
 | password          | string ?  | The account password.
-| type              | string ?  | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
+| type              | string    | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
 | ip                | string ?  | IP address associated with this account, used for auto-login.
 | team\_id          | ID ?      | The team that this account is for. Required iff type is `team`.
 | person\_id        | ID ?      | The person that this account is for, if the account is only for one person.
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -694,8 +694,8 @@ Properties of contest objects:
 | countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
-| scoreboard\_type             | string ?        | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
-| penalty\_time                | integer ?       | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
+| scoreboard\_type             | string          | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`. Defaults to `pass-fail`.
+| penalty\_time                | integer         | Penalty time for a wrong submission, in minutes. Only relevant if scoreboard\_type is `pass-fail`. Defaults to `0`.
```
eldering, line 698: I think both of these should simply be present if specified in `/access`: this is fundamental contest configuration, and there's only a single instance of each of these properties since they are in the contest object. So it should be whether this data is available and the server willing and able to provide it. And if the server doesn't specify these in `/access`, then there's still the default values that a client can rely on.

Contest_API.md
```diff
@@ -811,7 +811,7 @@ Properties of judgement type objects:
 | :------ | :-------- | :----------
 | id      | ID        | Identifier of the judgement type, a 2-3 letter capitalized shorthand, see table below.
 | name    | string    | Name of the judgement. (might not match table below, e.g. if localized).
-| penalty | boolean ? | Whether this judgement causes penalty time. Required iff contest:penalty\_time is present.
+| penalty | boolean   | Whether this judgement causes penalty time. Required iff contest:penalty\_time is present.
```
eldering, line 814: This is directly tied to the `contest.penalty_time` property. If that is present (and then presumed non-zero), then *each* judgement type should really have this defined for clients to be able to do anything sensible with this data. So I see no point in keeping this nullable.

Contest_API.md
```diff
@@ -916,11 +916,11 @@ Properties of language objects:
 | :------------------- | :--------------- | :----------
 | id                   | ID               | Identifier of the language from table below.
 | name                 | string           | Name of the language (might not match table below, e.g. if localized).
-| entry_point_required | boolean ?        | Whether the language requires an entry point. Defaults to `false`.
+| entry_point_required | boolean          | Whether the language requires an entry point. Defaults to `false`.
```
eldering, line 919: If this property is declared in `/access`, then I think it's reasonable to require that *each* language has it explicitly set (can still be either `false` or `true`).

Contest_API.md
```diff
@@ -916,11 +916,11 @@ Properties of language objects:
...
 | entry_point_name     | string ?         | The name of the type of entry point, such as "Main class" or "Main file"). Required iff entry_point_required is `true`.
 | extensions           | array of string  | File extensions for the language.
-| compiler             | Command object ? | Command used for compiling submissions.
-| runner               | Command object ? | Command used for running submissions. Relevant e.g. for interpreted languages and languages running on a VM.
+| compiler             | Command object   | Command used for compiling submissions.
+| runner               | Command object   | Command used for running submissions. Relevant e.g. for interpreted languages and languages running on a VM.
```
eldering, line 923: I think for both of these, if the server declared intent to provide these, then it should provide it for *each* language, thus these properties should not be nullable.

Contest_API.md
```diff
@@ -1032,9 +1032,9 @@ Properties of problem objects:
 | ordinal           | number    | A unique number that determines the order the problems, e.g. on the scoreboard.
 | rgb               | string ?  | Hexadecimal RGB value of problem color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet), e.g. `#AC00FF` or `#fff`.
 | color             | string ?  | Human readable color description associated to the RGB value.
-| time\_limit       | number ?  | Time limit in seconds per test data set (i.e. per single run). Should be an integer multiple of `0.001`.
-| test\_data\_count | integer ? | Number of test data sets.
-| max_score         | number ?  | Maximum expected score, although teams may score higher in some cases. Typically used to indicate scoreboard cell color in scoring contests. Required iff contest:scoreboard_type is `score`.
+| time\_limit       | number    | Time limit in seconds per test data set (i.e. per single run). Should be an integer multiple of `0.001`.
+| test\_data\_count | integer   | Number of test data sets.
```
eldering, line 1036: For both, if the server declares them in `/access`, then it really should be able to provide this information for *each* problem, so no point in allowing this to be nullable.

Contest_API.md
```diff
@@ -1032,9 +1032,9 @@ Properties of problem objects:
...
+| max_score         | number    | Maximum expected score, although teams may score higher in some cases. Typically used to indicate scoreboard cell color in scoring contests. Required iff contest:scoreboard_type is `score`.
```
eldering, line 1037: I guess this should be declared in `/access` only if (but not necessarily if) the scoreboard type is "score". If it is declared I guess it should be present for *each* problem, so not be nullable, but maybe @niemela can judge better.

Contest_API.md
```diff
@@ -1209,8 +1209,8 @@ Properties of team objects:
 | name              | string          | Name of the team.
 | display\_name     | string ?        | Display name of the team. If not set, a client should revert to using the name instead.
 | organization\_id  | ID ?            | Identifier of the [ organization](#organizations) (e.g. university or other entity) that this team is affiliated to.
-| group\_ids        | array of ID ?   | Identifiers of the [ group(s)](#groups) this team is part of (at ICPC WFs these are the super-regions). No meaning must be implied or inferred from the order of IDs. The array may be empty. Required iff groups endpoint is available.
-| hidden            | boolean ?       | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
+| group\_ids        | array of ID     | Identifiers of the [ group(s)](#groups) this team is part of (at ICPC WFs these are the super-regions). No meaning must be implied or inferred from the order of IDs. The array may be empty. Required iff groups endpoint is available.
```
eldering, line 1212: I think that if declared in `/access`, then this should at least be an empty array, but never `null`.

Contest_API.md
```diff
@@ -1209,8 +1209,8 @@ Properties of team objects:
...
+| hidden            | boolean         | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to false if missing.
```
eldering, line 1213: If declared in `/access`, then I see no burden in always providing this value.

Contest_API.md
```diff
@@ -1296,7 +1296,7 @@ Properties of account objects:
 | id                | ID        | Identifier of the account.
 | username          | string    | The account username.
 | password          | string ?  | The account password.
-| type              | string ?  | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
+| type              | string    | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
```
eldering, line 1299: Would there be any accounts that don't fall under one of these types (or even another not listed here)? If not, then I think this should be non-nullable.

Contest_API.md
```diff
@@ -1695,10 +1695,10 @@ Properties of clarification message objects:
 | Name           | Type    | Description
 | :------------- | :------ | :----------
 | id             | ID      | Identifier of the clarification.
-| from\_team\_id | ID ?    | Identifier of [ team](#teams) sending this clarification request, `null` if a clarification sent by jury.
-| to\_team\_id   | ID ?    | Identifier of the [ team](#teams) receiving this reply, `null` if a reply to all teams or a request sent by a team.
+| from\_team\_id | ID ?    | Identifier of [team](#teams) sending this clarification request, `null` iff a clarification sent by jury.
+| to\_team\_id   | ID ?    | Identifier of the [team](#teams) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
 | reply\_to\_id  | ID ?    | Identifier of clarification this is in response to, otherwise `null`.
-| problem\_id    | ID ?    | Identifier of associated [ problem](#problems), `null` if not associated to a problem.
+| problem\_id    | ID ?    | Identifier of associated [problem](#problems), `null` iff not associated to a problem.
```
eldering, line 1701: Make it explicit that these attributes can *only* be null under the specified conditions.

Contest_API.md
```diff
@@ -1971,14 +1971,14 @@ The following endpoints are associated with commentary:
...
-| team\_ids     | array of ID ? | JSON array of [team](#teams) IDs the message is related to.
-| problem\_ids  | array of ID ? | JSON array of [problem](#problems) IDs the message is related to.
+| Name          | Type        | Description
+| :------------ | :---------- | :----------
+| id            | ID          | Identifier of the commentary.
+| time          | TIME        | Time of the commentary message.
+| contest\_time | RELTIME     | Contest time of the commentary message.
+| message       | string      | Commentary message text. May contain special tags for [teams](#teams) and [problems](#problems) on the format `#t<team ID>` and `#p<problem ID>` respectively.
+| team\_ids     | array of ID | JSON array of [team](#teams) IDs the message is related to.
+| problem\_ids  | array of ID | JSON array of [problem](#problems) IDs the message is related to.
```
eldering, line 1981: If these team/problem IDs properties are listed in `/access`, then I think they should always contain an empty array instead of `null`.

Contest_API.md
```diff
@@ -2064,7 +2064,7 @@ Properties of problem data objects:
 | problem\_id  | ID        | Identifier of the [ problem](#problems).
 | num\_judged  | integer   | Number of judged submissions (up to and including the first correct one),
 | num\_pending | integer   | Number of pending submissions (either queued or due to freeze).
-| solved       | boolean ? | Required iff contest:scoreboard_type is `pass-fail`.
+| solved       | boolean   | Required iff contest:scoreboard_type is `pass-fail`.
```
eldering, line 2067: If the scoreboard type is "pass-fail", then this always has a well-defined value, so let's be explicit about it and not allow `null`.

Contest_API.md
niemela: I'm torn about this one. Missing (i.e. null) instead of false is a nice shorthand?

Contest_API.md
niemela: Could there be individual problems where one or both of these don't make sense?

Contest_API.md
niemela, line 1213:
```suggestion
| hidden            | boolean         | If the team is to be excluded from the [scoreboard](#scoreboard). Defaults to `false`.
```

Contest_API.md
niemela: Since it's just a string and not an "enum" (compare with people:sex) it would certainly fall under *some* possible value. If nothing else `other` would do.

Contest_API.md
niemela: Given the vagueness built-in ("although teams may score higher in some cases") I think we could make this non-nullable.

Contest_API.md
niemela: I'm torn about this one.

Contest_API.md
niemela: But it's currently not possible to state in `/access` that you are not providing this.

Contest_API.md
eldering: I'd really prefer to be explicit instead of allowing shorthand alternatives. If you already declared this property in `/access`, then you most likely have some languages for which `entry_point_required=true`, so I don't think it saves much to be able to leave the attribute out, and that it doesn't weigh up to the clarity of explicitly having it present.

Contest_API.md
eldering: That is a good point, but let's try to see if we can add that. It feels a bit weird that here and also for `scoreboard.rows.problems` we have it nullable, just because we can't specify it in `/access`.

niemela: Agreed.

Contest_API.md
nickygerritsen: I agree with Jaap. I also can't think of an implementation (either server or client) that would be easier if we do this. I think it only saves bytes on the wire, which should not be a reason IMHO.

Contest_API.md
nickygerritsen: Yeah I think `other` is fine? For us for example we have accounts that we otherwise should not expose (like a `CDS`or `kattis` user). Or should we really not expose them? 🤔

Contest_API.md
nickygerritsen: I think this is the same as for the entry point thing: it doesn't really save a lot IMHO to make this nullable.

nickygerritsen (commented):
One thing I noticed earlier, which is somewhat related:

For `frozen` and `thawed` in the state endpoint we say something like:

> Required iff scoreboard_freeze_duration is present in the [contest](https://ccs-specs.icpc.io/master/contest_api#contests) endpoint.

But the freeze duration can be present but `0:00:00` which is the same as it not being present. We should maybe reword this?

Contest_API.md
niemela: Ok, yeah, those are good points. I agree.

niemela:
> But the freeze duration can be present but `0:00:00` which is the same as it not being present. We should maybe reword this?

No, I don't think we should. On one hand setting free duration to 0 is effectively equivalent with having no freeze, except you have to thaw. Which kinda makes sense, because technically, you do have a freeze. More importantly though, I don't think we need to solve the "problem" that somebody wants to declare a freeze duration of 0 but does not want to thaw. The solution to that is to *not* have a freeze.

niemela (dismissed):

nickygerritsen:
> > But the freeze duration can be present but `0:00:00` which is the same as it not being present. We should maybe reword this?
> 
> No, I don't think we should. On one hand setting free duration to 0 is effectively equivalent with having no freeze, except you have to thaw. Which kinda makes sense, because technically, you do have a freeze. More importantly though, I don't think we need to solve the "problem" that somebody wants to declare a freeze duration of 0 but does not want to thaw. The solution to that is to _not_ have a freeze.

If I understand correctly the only way to have no freeze is to not put it in access, right? Otherwise it defaults to 0, meaning you have a freeze.

so we have foue cases:
* there is no freeze. You don’t have to set frozen/thawed
* the freeze  is 0 but you didn’t supply it. You don’t have to set frozen/thawed
* the freeze is 0 and you set it. You have to set the fields
* the freeze is non zero. You have to set the fields

I don’t think this is what we want (especially point 2) but you seem to disagree?

niemela:
> If I understand correctly the only way to have no freeze is to not put it in access, right? Otherwise it defaults to 0, meaning you have a freeze.
> 
> so we have foue cases:
> 
>     * there is no freeze. You don’t have to set frozen/thawed
> 
>     * the freeze  is 0 but you didn’t supply it. You don’t have to set frozen/thawed
> 
>     * the freeze is 0 and you set it. You have to set the fields
> 
>     * the freeze is non zero. You have to set the fields
> 
> 
> I don’t think this is what we want (especially point 2) but you seem to disagree?

I agree that we don't want that, I disagree that this is what we have.

- If `scoreboard_freeze_duration` is **not** in `/access` you are **not** allowed to provide it in `/contests/<id>`, and you **don't** have to provide `frozen` and `thawed` in `/state`. This should probably be changed to explicitly say that you *must* not provide  `frozen` and `thawed`, but OTOH it's kinda obvious.
- If `scoreboard_freeze_duration` **is**  in `/access` you and you **don't** provide it in `/contests/<id>` (or explicitly set it to `null`, which must mean the same thing), then you **don't** have to provide `frozen` and `thawed` in `/state`.
- If `scoreboard_freeze_duration` **is**  in `/access` you and you **do** provide any value (`null` is not a value) in `/contests/<id>`, including `0:00:00`, then you **do* have to provide `frozen` and `thawed` in `/state`.

Explicitly providing the default value (`0:00:00`) will just force you to provide `frozen` and `thawed`. If you don't want that, then don't do that.

Contest_API.md
nickygerritsen: I get your point, this means you need to somehow know your version information before you can expose languages, which might mean you need to connect to some judgehost hardware to figure that out.

But you could of course just not send out the languages until then. Why should you not be able to import CMS data?

deboer-tim:
My ISP just went down for 45 minutes so there is at least one case above where Fredrik appears to be agreeing to my point when I am assuming he was actually replying to the previous comment. :)

I'm about to send an email with my thoughts, but to Nicky's point above I think we're over-complicating things by making /access synonymous with both knowing about a property and having a value. I think there can be cases where something exists but I don't have data yet, or 'I know about scoring contests, and this isn't one'. Maybe these aren't ideal cases, but I think we're making things harder to implement by forcing them.

nickygerritsen:
> > If I understand correctly the only way to have no freeze is to not put it in access, right? Otherwise it defaults to 0, meaning you have a freeze.
> > so we have foue cases:
> > ```
> > * there is no freeze. You don’t have to set frozen/thawed
> > 
> > * the freeze  is 0 but you didn’t supply it. You don’t have to set frozen/thawed
> > 
> > * the freeze is 0 and you set it. You have to set the fields
> > 
> > * the freeze is non zero. You have to set the fields
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > I don’t think this is what we want (especially point 2) but you seem to disagree?
> 
> I agree that we don't want that, I disagree that this is what we have.
> 
> * If `scoreboard_freeze_duration` is **not** in `/access` you are **not** allowed to provide it in `/contests/<id>`, and you **don't** have to provide `frozen` and `thawed` in `/state`. This should probably be changed to explicitly say that you _must_ not provide  `frozen` and `thawed`, but OTOH it's kinda obvious.
> * If `scoreboard_freeze_duration` **is**  in `/access` you and you **don't** provide it in `/contests/<id>` (or explicitly set it to `null`, which must mean the same thing), then you **don't** have to provide `frozen` and `thawed` in `/state`.
> * If `scoreboard_freeze_duration` **is**  in `/access` you and you **do** provide any value (`null` is not a value) in `/contests/<id>`, including `0:00:00`, then you *_do_ have to provide `frozen` and `thawed` in `/state`.
> 
> Explicitly providing the default value (`0:00:00`) will just force you to provide `frozen` and `thawed`. If you don't want that, then don't do that.

OK I can agree to this. I still find it kinda strange there is a different between explicitly specifying a default value somewhere and specifying `null` (which defaults to...the default value) but it's a case that should not happen so 🤷 .

Contest_API.md
niemela: > I get your point, this means you need to somehow know your version information before you can expose languages

No, you only need to know the `command`, all other fields are optional. And you *could* always put in a dummy value. It's exactly equally useless.

nickygerritsen: Ah sorry, didn't see those fields were optional. Then yes, I think you'd know your command at the point you load the language, right? And if not put in something and change it later indeed.

niemela:
> OK I can agree to this. I still find it kinda strange there is a different between explicitly specifying a default value somewhere and specifying `null` (which defaults to...the default value) but it's a case that should not happen so 🤷 .

Yeah, it is a bit strange. Would changing "if present" to "if not 0" be better? I'm not sure. Is that the change you originally wanted?

But I also don't think it's a big problem.

nickygerritsen:
> > OK I can agree to this. I still find it kinda strange there is a different between explicitly specifying a default value somewhere and specifying `null` (which defaults to...the default value) but it's a case that should not happen so 🤷 .
> 
> Yeah, it is a bit strange. Would changing "if present" to "if not 0" be better? I'm not sure. Is that the change you originally wanted?
> 
> But I also don't think it's a big problem.

Yeah I'm fine with leaving it. Putting `if not 0` would also confuse people.

Contest_API.md
deboer-tim: > I don't agree with the "if it's not in `/access` id doesn't exists and if it doesn't exists it can't have a default".
> 
> We agree that you should be allowed to not provide `scoreboard_type`, right? And that that means that scoreboard is `pass-fail`? This is based on the non-existing property having a default.
>
> Furthermore, any future extension surely currently does not exist. And you would definitely want to provide a default for that.

Not having penalty_time in /access doesn't mean that there is a penalty time and it is 0, there just _isn't_ a penalty time. I totally agree that we need to define the _behaviour_ when a property doesn't exist, it's just confusing to me to specify a default for a property that we've said can't have one. Using language like "If it doesn't exist, ..." would make more sense to me.

Contest_API.md
deboer-tim: Sure, you can enter 'something' in any non-nullable property - but why require anyone do that, require provider to not send events until specific properties are added, or require manual interaction importing data from CMS? To me these are all additional requirements to solve a problem that doesn't exist. Nobody ever said 'don't tell me anything about a problem until it's 100% configured' or 'you should be forced to make up data and fix it later'. CCS can still decide to do that, but IMHO shouldn't be spec-required.

Contest_API.md
deboer-tim: I honestly think there are lots of cases where 'it wasn't provided to us' or 'we haven't configured that yet' (not necessarily just this property) and that's exactly _why_ we should keep properties like this nullable (or just missing/not set yet). We've had lots of data sets over the years that are incomplete or take time to fix, and there's a difference between 'it would be dumb to start a contest like this' and 'this state shouldn't be allowed to ever exist and your system must be built to enforce it'.

Contest_API.md
eldering: The way I see it: if you already have languages, but somehow don't have the compiler/runner command data yet, then you can simply not add them to `/access` and not provide these properties. Only once you have them properly configured/loaded, you add them to `/access` and provide them in the languages endpoint.

Contest_API.md
eldering: _Maybe_ there are problems where a test data count or timelimit doesn't make sense, but I honestly cannot think of one. Also, I think `test_data_count = 0` is a perfectly reasonably response that provides more information than `null`.

If we don't have this information available intially or for certain problems, then I think that was/is a bug at that time and needs fixing, so the fact that it then is not according to spec, I see as a correct reflection of what's happening. In other words, if I'd then run a validation check against the API, I want to see this get flagged.

Contest_API.md
eldering: > One thing I noticed earlier, which is somewhat related:
> 
> For `frozen` and `thawed` in the state endpoint we say something like:
> 
> > Required iff scoreboard_freeze_duration is present in the [contest](https://ccs-specs.icpc.io/master/contest_api#contests) endpoint.
> 
> But the freeze duration can be present but `0:00:00` which is the same as it not being present. We should maybe reword this?

Agreed. I think we should reword that part so that the value `0:00:00` is explicitly forbidden and `null` means "no freeze".

Contest_API.md
```diff
@@ -689,13 +689,13 @@ Properties of contest objects:
 | :--------------------------- | :-------------- | :----------
 | id                           | ID              | Identifier of the current contest.
 | name                         | string          | Short display name of the contest.
-| formal\_name                 | string          | Full name of the contest.
+| formal\_name                 | string ?        | Full name of the contest. Defaults to value of `name`.
```
eldering, line 692: I'd still argue that this doesn't have to be nullable, since there's only a single instance of this in the contest object, and thus the server can simply indicate in `/access` whether this will be available. But I can live with this.

deboer-tim: Ditto, I'd be ok with this not being nullable.

deboer-tim (approved):

niemela (approved):

PR #96, opened 2022-03-29
Change CCS system requirements to require import from a contest archive.

nickygerritsen:
I removed a lot of stuff and just said 'you need to be able to import x, y, z from a contest archive'.

Does that make sense?

I think it does, but I do have some questions:
* Are the fields I specify actually what we want? For example the logo field for organizations and the password field for accounts are not in the required access fields, but you do need to be able to import them. I think that makes sense.

* Currently we say "if you have username/password auth, you need to be able to import `accounts.json`". But later we say "you NEED to be able to import `accounts.json`". This is inconsistent. Should we just require being able to import it?

* I remember vividly us talking about allowing YAML support for `contest.json` and `problems.json` and that is also in the agenda doc (with agreement). But it's nowhere to be found. How do we want to word this? I could add a line "For `contest.json` and `problems.json` YAML files with equivalent content is also allowed" but I'm not sure if makes sense?

* I would propose we also allow YAML for `accounts.json`. This is, together with the contest and problem files, one that is written by a human quite often and (simple) YAML is way easier to write as a human.

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -159,13 +159,7 @@ were some user other than that specified by their login credentials.
 
 If the CCS uses a login/password mechanism to enforce secure
 authentication, it must support account creation and password assignment
-according to the following algorithm:
-
-1.  read [teams.tsv](#teamstsv) defining team accounts to be
-    created
-2.  read [accounts.tsv](#accountstsv) defining additional
-    accounts to be created or, for teams, specifying the password of the
-    accounts.
+by reading [accounts.json](#importing-contest-configuration).
 
 #### Logging Out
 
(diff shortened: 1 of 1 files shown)


deboer-tim:
> * Are the fields I specify actually what we want? For example the logo field for organizations and the password field for accounts are not in the required access fields, but you do need to be able to import them. I think that makes sense.

It was a little odd before that we required the CCS to use provided logos from a gzip (which we never really did) but only exposed them on its scoreboard and not via API. Technically we still don't need CCS logos in API at finals since the CDS will provide if the CCS doesn't. There's also optional data in the CAF (e.g. locations) that the CCS isn't required to support.

To me the current "required properties are X" and "you must read everything required from a CAF" is already better than what we have.

> * Currently we say "if you have username/password auth, you need to be able to import `accounts.json`". But later we say "you NEED to be able to import `accounts.json`". This is inconsistent. Should we just require being able to import it?

Yeah, I think so. CCS needs team accounts at a minimum, this is where they're defined.

> * I remember vividly us talking about allowing YAML support for `contest.json` and `problems.json` and that is also in the agenda doc (with agreement). But it's nowhere to be found. How do we want to word this? I could add a line "For `contest.json` and `problems.json` YAML files with equivalent content is also allowed" but I'm not sure if makes sense?

Yes, I think we ended up with 'both should be supported'. Should we add these as options in the CAF, or just in the CCS spec? Either way, I think we need to keep examples of these files since the json-via-yaml conversion might not be obvious.

> * I would propose we also allow YAML for `accounts.json`. This is, together with the contest and problem files, one that is written by a human quite often and (simple) YAML is way easier to write as a human.

Agreed.

Contest_Control_System_Requirements.md
```diff
@@ -245,30 +239,23 @@ be required to make any changes once the contest starts.
 
 ### Importing Contest Configuration
 
-The CCS must be able to import contest configuration from a single URL,
+The CCS must be able to import contest configuration from a
+[Contest Archive](contest_archive_format),
 and use the data at that location to configure the CCS. The
```
eldering, line 244:
```suggestion
and use the data to configure the CCS. The
```

Contest_Control_System_Requirements.md
```diff
@@ -245,30 +239,23 @@ be required to make any changes once the contest starts.
...
-<http://10.1.1.0/data>, teams.tsv can be imported from
-<http://10.1.1.0/data/teams.tsv>.
-
-The CCS must be able to read [teams.tsv](#teamstsv), which
-contains a list of teams registered for the World Finals. It must use
-the data in this file to automatically create whatever internal
-representation is necessary to allow each team (and only the teams)
-specified in the registration list to participate in the World Finals
-contest.
+configuration URL contains all contest configuration data, including
```
eldering, line 245:
```suggestion
archive contains all contest configuration data, including
```

Contest_Control_System_Requirements.md
```diff
@@ -348,7 +335,7 @@ The CCS must allow updating of configuration data without restarting or
 stopping the CCS.
 
 The CCS must support the ability to create and save an updated
-contest.yaml file upon request.
+contest.json file upon request.
```
niemela, line 338: This follows from the Contest API requirement. I'd suggest dropping this.

Contest_Control_System_Requirements.md
```diff
@@ -1020,6 +1007,12 @@ The CCS must provice access to the Contest API including the
             "group_ids"
          ]
       },
+      { 
+         "type": "accounts", 
+         "properties": [
+            "id", "username", "type", "ip", "team_id"
+         ]
+      },
```
niemela, line 1015: I don't think we need to require the CCS to be able to export accounts.

Even if we do, I don't think we need to include "ip".

nickygerritsen: I’m fine with dropping this. I would like to add IP to the import of you all agree

nickygerritsen:
OK I fixed the comments. The YAML part needs attention, I'm not sure if this is how we want to do it.

Contest_Control_System_Requirements.md
niemela: What would that mean though? There is no requirement that requires the use of the information, so what does it mean that it must be imported?

nickygerritsen: Are you explicitly referring to IP? I would like to add that as a requirement somewhere, but not sure where. I'm also fine not doing that though.

deboer-tim (approved):

PR #99, opened 2022-04-27
Reconnection tokens

deboer-tim:
As discussed on calls, the event feed reconnection was left incomplete after changing notifications and adding webhooks, and this is a path to fix it.

Many events are outside of or don't include contest time, the client and server may not agree on system time, and I don't like scraping events to find specific time properties to identify them. The only other options is to add a new property, which I've called 'token'. The token is optional and could be anything (id, counter, timestamp, etc). The allowed support options are pretty wide and some options are listed in a new use cases section.


nickygerritsen (approved):
I like the idea. I **do** think we should make the token and since_token stuff required for the WF CCS system requirements, since otherwise it seems to me that reconnecting can be super slow.

PR #103, opened 2022-05-30
Update tag format for commentary. Add priority and source.

niemela:


Contest_API.md
```diff
@@ -1987,11 +1987,13 @@ Properties of commentary objects:
 | id            | ID            | Identifier of the commentary.
 | time          | TIME          | Time of the commentary message.
 | contest\_time | RELTIME       | Contest time of the commentary message.
-| message       | string        | Commentary message text. May contain special tags for [teams](#teams) and [problems](#problems) on the format `#t<team ID>` and `#p<problem ID>` respectively.
+| message       | string        | Commentary message text. May contain special tags refering to endpoint objects on the format `{<endpoint>:<object ID>}`. E.g. this is most commonly used for references to [teams](#teams) and [problems](#problems) as `{teams:<team ID>}` and `{problems:<problem ID>}` respectively.
+| priority      | number        | Importance of the commentary message. Higher value means more important. Intended to be used to filter messages.
+| source        | ID ?          | Source [person](#persons) of the commentary message.
```
eldering, line 1992: I think we should say something about the range of `priority`. Is `0` the lowest, or the default and can it go negative? What would be "top priority"?

Contest_API.md
```diff
@@ -1987,11 +1987,13 @@ Properties of commentary objects:
...
-| message       | string        | Commentary message text. May contain special tags for [teams](#teams) and [problems](#problems) on the format `#t<team ID>` and `#p<problem ID>` respectively.
+| message       | string        | Commentary message text. May contain special tags refering to endpoint objects on the format `{<endpoint>:<object ID>}`. E.g. this is most commonly used for references to [teams](#teams) and [problems](#problems) as `{teams:<team ID>}` and `{problems:<problem ID>}` respectively.
+| priority      | number        | Importance of the commentary message. Higher value means more important. Intended to be used to filter messages.
+| source        | ID ?          | Source [person](#persons) of the commentary message. 
 | team\_ids     | array of ID ? | JSON array of [team](#teams) IDs the message is related to.
 | problem\_ids  | array of ID ? | JSON array of [problem](#problems) IDs the message is related to.
 
-For the message, if an literal `#` is needed, `\#` must be used. Similarly for literal `\`, `\\` must be used.
+For the message, if an literal `{` is needed, `\{` must be used. Similarly for literal `\`, `\\` must be used.
```
eldering, line 1996: Small typo fix

```suggestion
For the message, if a literal `{` is needed, `\{` must be used. Similarly for literal `\`, `\\` must be used.
```

nickygerritsen (dismissed):
One comment. Looks good to me otherwise

nickygerritsen (dismissed):

deboer-tim (changes requested):
`source` should be `source_id` for consistency with other endpoints (e.g. clarification from_team_id).

Maybe a tag for human or automated commenter? Or do we assume that the source will be set for humans so we don't need a tag?

niemela:
> `source` should be `source_id` for consistency with other endpoints (e.g. clarification from_team_id).

For some reason I really don't like this. It's not a *source* ID, it's a *person* ID (of the person that is the source).

Also, why do we have to tag IDs and only IDs with `_id`? Why not `message_str` because it's a string, and so on...?

That said, I have done it :(.

> Maybe a tag for human or automated commenter? Or do we assume that the source will be set for humans so we don't need a tag?

Let's see if it's needed. If it is we add it later, ok?

deboer-tim (approved):

eldering:
> > > `source` should be `source_id` for consistency with other endpoints (e.g. clarification from_team_id).
> > 
> > 
> > For some reason I really don't like this. It's not a _source_ ID, it's a _person_ ID (of the person that is the source).
> 
> Maybe it should be person_id or author_id then? Yeah, I know none of these are ideal.
> 
> > Also, why do we have to tag IDs and only IDs with `_id`? Why not `message_str` because it's a string, and so on...?
> > That said, I have done it :(.
> 
> lol. Because id is special! (but mostly because it became a pattern)

I'd say to make it explicit that it references another collection/endpoint.

Contest_API.md
```diff
@@ -1982,16 +1982,43 @@ The following endpoints are associated with commentary:
...
-| team\_ids     | array of ID ? | JSON array of [team](#teams) IDs the message is related to.
-| problem\_ids  | array of ID ? | JSON array of [problem](#problems) IDs the message is related to.
-
-For the message, if an literal `#` is needed, `\#` must be used. Similarly for literal `\`, `\\` must be used.
+| Name          | Type            | Description
+| :------------ | :-------------- | :----------
+| id            | ID              | Identifier of the commentary.
+| time          | TIME            | Time of the commentary message.
+| contest\_time | RELTIME         | Contest time of the commentary message.
+| message       | string          | Commentary message text. May contain special tags refering to endpoint objects on the format `{<endpoint>:<object ID>}`. E.g. this is most commonly used for references to [teams](#teams) and [problems](#problems) as `{teams:<team ID>}` and `{problems:<problem ID>}` respectively.
```
eldering, line 1990:
```suggestion
| message       | string          | Commentary message text. May contain special tags referring to endpoint objects in the format `{<endpoint>:<object ID>}`. This is most commonly used for references to [teams](#teams) and [problems](#problems) as `{teams:<team ID>}` and `{problems:<problem ID>}` respectively.
```

Contest_API.md
```diff
@@ -1982,16 +1982,43 @@ The following endpoints are associated with commentary:
...
+| tags          | array of string | JSON array of tags describing the message.
+| source_id     | ID ?            | Source [person](#persons) of the commentary message.
+| team\_ids     | array of ID ?   | JSON array of [team](#teams) IDs the message is related to.
+| problem\_ids  | array of ID ?   | JSON array of [problem](#problems) IDs the message is related to.
+
+For the message, if a literal `{` is needed, `\{` must be used. Similarly for literal `\`, `\\` must be used.
```
eldering, line 1996:
```suggestion
For the message, if a literal `{` is needed, `\{` must be used. Similarly for a literal `\`, `\\` must be used.
```

Contest_API.md
niemela: Huh... is it really "in the format" and not "on the format"?

Contest_API.md
```diff
@@ -1982,16 +1982,43 @@ The following endpoints are associated with commentary:
...
-
-For the message, if an literal `#` is needed, `\#` must be used. Similarly for literal `\`, `\\` must be used.
+| Name          | Type            | Description
+| :------------ | :-------------- | :----------
+| id            | ID              | Identifier of the commentary.
+| time          | TIME            | Time of the commentary message.
+| contest\_time | RELTIME         | Contest time of the commentary message.
+| message       | string          | Commentary message text. May contain special tags refering to endpoint objects on the format `{<endpoint>:<object ID>}`. E.g. this is most commonly used for references to [teams](#teams) and [problems](#problems) as `{teams:<team ID>}` and `{problems:<problem ID>}` respectively.
+| tags          | array of string | JSON array of tags describing the message.
+| source_id     | ID ?            | Source [person](#persons) of the commentary message.
```
eldering, line 1992: I didn't follow the previous discussion in detail, but if this is really a *person* (and not an automated system), then wouldn't `author_id` be a better name?

Contest_API.md
eldering: I _think_ so, but I'm not a native speaker, so maybe @deboer-tim should decide :-)

deboer-tim: Not sure I should be calling the shots given my grammar some days. :) IMHO 'in' is better than 'on', could also use 'with' or 'using'. Maybe 'using' is best?

Contest_API.md
deboer-tim: When this is set (many contests don't have person info) it would refer to the person who wrote it. Based on how we did clarifications this would be `from_person_id`, but `author_id` is probably the best of these three options.

niemela: ...but it certainly *could* be a system.

Now, if it's only one system, and especially if it's "the system" (whatever that means) it would make sense to use `null`. E.g. this would be reasonable as source for all the automatic commentary that Kattis generates.

But what if we have multiple systems pushing up updates to commentary?

Contest_API.md
niemela: Fixed (with "using").

Contest_API.md
niemela: Fixed.

Contest_API.md
eldering: I'd be in favour to keep it simple and use `null` for system. Representing systems as persons sounds weird.

I think I also like `from_person_id` a bit better to be consistent with clarifications

niemela: I don't think `from_person_id` is better. I suggest merging now as is, and we can discuss changes on the next meeting.

nickygerritsen (approved):

PR #105, opened 2022-06-20
Multiple systems as subfolders

deboer-tim:
As discussed on the CLICS call I rewrote the section on multiple systems to change from using a naming postfix to separate subfolders. I moved it to its own section so that it could be expanded into multiple (short) paragraphs, and I tried to explain the reasoning a bit more.

I went with the unimaginative folder name of 'systems', feel free to suggest better.

### DIFF
--- a/Contest_Archive_Format.md
+++ b/Contest_Archive_Format.md
@@ -91,7 +91,7 @@ archives of any of the other systems.
 
 When there is duplicate data (or the differences between systems are
 irrelevant), a system's data can be deleted and replaced with symlinks to
-the data in the parent folder. Using this method reduces the overall size
+the data from the primary system. Using this method reduces the overall size
 of the archive while still maintaining a valid contest archive (the system's
 view of the contest) for each system.
 
--- a/Contest_Archive_Format.md
+++ b/Contest_Archive_Format.md
@@ -73,15 +73,27 @@ Optionally one could create a Shallow Archive by not storing the files, in
 which case the URLs must be valid. This could be useful in some cases where
 the size of the archive matters.
 
-In some cases it could make sense to merge multiple API sources (of the same
-contest) in a single archive. One example of this would be a contest that was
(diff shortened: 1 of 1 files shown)


Contest_Archive_Format.md
```diff
@@ -73,15 +73,27 @@ Optionally one could create a Shallow Archive by not storing the files, in
...
+be stored as described in this document. Each additional system should
+store its data using the same format but in a unique sub-folder
+`systems/<system>`, named after the system that produced it.
+Using this pattern means that tools that work against a contest archive
+can be run without change on either the primary contest archive or the
+archives of any of the other systems.
+
+When there is duplicate data (or the differences between systems are
+irrelevant), a system's data can be deleted and replaced with symlinks to
+the data in the parent folder. Using this method reduces the overall size
```
eldering, line 94: Nitpick: `parent` is not exactly correct, so I'd suggest:
```suggestion
the data from the primary system. Using this method reduces the overall size
```

Contest_Archive_Format.md
```diff
@@ -73,15 +73,27 @@ Optionally one could create a Shallow Archive by not storing the files, in
...
+Some contests are run with multiple systems that provide a Contest API.
+One example of this is a contest run with a primary and
+[shadow](ccs_system_requirements#shadow-mode) CCS. Typically in these cases
+there is relationship between the systems or a lot of common data, so it
+makes sense to merge the data into a single archive.
+
+Data from the primary system (usually the primary CCS) should continue to
+be stored as described in this document. Each additional system should
+store its data using the same format but in a unique sub-folder
+`systems/<system>`, named after the system that produced it.
```
eldering, line 87: Maybe `other-systems`? I'm not sure it's better, but it does make it more explicit that the root directory already contains a full archive and that these are additional ones.

deboer-tim: I don't love this, but the only downside I could come up with 'it's longer' so I made the change. :)

PR #108, opened 2022-07-25
Reconnection tokens

deboer-tim:
As discussed on today's call, this is an optional mechanism for servers to provide short-term reconnections for clients that request it.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -2174,7 +2174,7 @@ the same order or set of events. The only guarantees are:
 - the latest notification sent for any object is the correct and current
 state of that object. E.g. if an object was created and deleted the
 delete notification will be sent last.
-- when a notification is sent the change it decsribes must already have
+- when a notification is sent the change it describes must already have
 happened. I.e. if a client receives an update for a certain endpoint a
 `GET` from that endpoint will return that state or possible some later
 state, but never an earlier state.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -2191,8 +2191,9 @@ specify the last notification token it received:
 
 If specified, the server will attempt to start sending events since the
 given token to reduce the volume of events and required reconciliation.
-If the token is invalid, the time passed is too large (typically no more
-than 15 minutes), or the server does not support this parameter, the
+If the token is invalid, the time passed is too large (a server that
+supports since_token should support an expiry time of at least
+15 minutes), or the server does not support this parameter, the
 request will fail with a 400 error.
 
 The client is guaranteed to either get a 400 error or receive at
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -2189,16 +2189,16 @@ specify the last notification token it received:
 
 `/event-feed?since_token=xx`
 
-If specified, the server will attempt to start sending events around the
+If specified, the server will attempt to start sending events since the
 given token to reduce the volume of events and required reconciliation.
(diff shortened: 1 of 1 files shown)


niemela (dismissed):

nickygerritsen (dismissed):

Contest_API.md
```diff
@@ -2183,17 +2184,34 @@ state, but never an earlier state.
...
-occurred during the time the client was disconnected will be reflected.
+given token to reduce the volume of events and required reconciliation.
+If the token is invalid, the time passed is too large (typically no more
+than 15 minutes), or the server does not support this parameter, the
+request will fail with a 400 error.
+
+There is no guarantee that all updates (e.g. a team name correction,
+which is not time-based) that occurred during the time the client was
+disconnected will be reflected. The server may also resend some
+notifications to ensure that they were received.
```
eldering, line 2201: So changes in static data (like a team name change) are tricky, but I think it should be guaranteed at least that the server will provide all updates to live data required to get into a consistent state again with reality.

The reason is that also the "start sending events _around_" is a bit vague. Currently it doesn't seem guaranteed that the server doesn't just skip a few events after `since_token`.

niemela: Yeah, I think we need a *little* bit stronger guarantees.

How about something like: if `since_token` is used the client will either get a 400 *or* it is guaranteed that it will get at least all changes since the token (but it *could* also get a lot more).

deboer-tim: Please review again. I removed the word 'around' and used Fredrik's text. I felt it needed a reason why it might receive more events so included that as well.

nickygerritsen (approved):

deboer-tim:
Another review please @eldering or @niemela ?

niemela (dismissed):

Contest_API.md
```diff
@@ -2183,17 +2184,34 @@ state, but never an earlier state.
...
+If the token is invalid, the time passed is too large (typically no more
+than 15 minutes), or the server does not support this parameter, the
+request will fail with a 400 error.
+
+The client is guaranteed to either get a 400 error or receive at
+least all changes since the token (but it could also get a lot more,
+e.g. if the server is multithreaded and needs to resend some events
+to ensure they were received).
+
+###### Use cases for reconnecting
```
eldering, line 2203:
```suggestion
###### Server support for reconnecting
```
I think this describes the section below better.

Contest_API.md
```diff
@@ -2183,17 +2184,34 @@ state, but never an earlier state.
...
-If specified, the server will attempt to start sending events around the
-given time to reduce the volume of events and required reconciliation.
-If the time passed is too large or the server does not support this
-parameter, all objects will be sent. There is no guarantee that all
-updates (e.g. a team name correction, which is not time-based) that
-occurred during the time the client was disconnected will be reflected.
+If specified, the server will attempt to start sending events since the
+given token to reduce the volume of events and required reconciliation.
+If the token is invalid, the time passed is too large (typically no more
+than 15 minutes), or the server does not support this parameter, the
```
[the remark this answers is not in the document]
deboer-tim: I changed the line in brackets from a client-side expectation to the server-side requirement you suggested. Does that work?

nickygerritsen (approved):

niemela (approved):

PR #110, opened 2022-08-16
Add changes since previous version.

nickygerritsen:


niemela (approved):

deboer-tim (approved):
We also added problem package and statement (although I'm good merging as-is and adding more later).

FWIW, I think we should go back and add one of these for the previous version too, e.g. to note the addition of scoring contests, SVGs, group scoreboards, etc.

deboer-tim (approved):

niemela (approved):

PR #111, opened 2022-08-16
Add changes since previous version.

nickygerritsen:


niemela:
My apologies if I missed a discussion somewhere, but why do we want to merge changes since previous version (which is 2021-11) into 2021-11? @nickygerritsen

nickygerritsen:
> My apologies if I missed a discussion somewhere, but why do we want to merge changes since previous version (which is 2021-11) into 2021-11? @nickygerritsen

I'm confused now. This should be the changes between `2020-03` and `2021-11`. We never did this and `2021-11` was some 'in between' version used only at NWERC in Iceland to test stuff, so I think it makes sense to tell people what changed since the first released version since we release versions in this format.

niemela:
@nickygerritsen Sorry... I was confused. Disregard...

niemela (approved):

PR #120, opened 2022-09-20
Person Clarifications

deboer-tim:
Minor clarifications for persons:
- Role didn't match the possible roles, change it to be for coaches and contestants.
- Add 'other' role to account for people that aren't staff or directly on a team, e.g. spare team members, attendees, or local officials that are in CMS.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1271,7 +1271,7 @@ Properties of person objects:
 | :---------- | :-------------- | :----------
 | id          | ID              | Identifier of the person.
 | icpc\_id    | string ?        | External identifier from ICPC CMS.
-| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `contestant` or `coach`.
+| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `contestant`.
 | name        | string          | Name of the person.
 | title       | string ?        | Title of the person, e.g. "Technical director".
 | email       | string ?        | Email of the person.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1271,12 +1271,12 @@ Properties of person objects:
 | :---------- | :-------------- | :----------
 | id          | ID              | Identifier of the person.
 | icpc\_id    | string ?        | External identifier from ICPC CMS.
-| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `team`.
+| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `contestant` or `coach`.
 | name        | string          | Name of the person.
 | title       | string ?        | Title of the person, e.g. "Technical director".
 | email       | string ?        | Email of the person.
 | sex         | string ?        | Either `male` or `female`, or possibly `null`.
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1271,12 +1271,12 @@ Properties of person objects:
 | :---------- | :-------------- | :----------
 | id          | ID              | Identifier of the person.
 | icpc\_id    | string ?        | External identifier from ICPC CMS.
-| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `team`.
+| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `contestant` or `coach`.
```
eldering, line 1274: What if a coach is coach for multiple teams? I think that's possible (at least at NWERC), while I suppose a contestant must always be member exactly 1 team.

nickygerritsen: It is. We discussed this and this will be for the next draft instead of the current spec since it's a more backwards incompatible change.

nickygerritsen (approved):

Contest_API.md
eldering: But should we then make it required for coaches now? I think it then makes more sense to leave it open and up to the system/admin to decide what to do.

deboer-tim: Not sure I understand - you think we should leave it ok for a coach to not be associated with any team until they can be associated with more than one? I guess you'd prefer not to have to pick one team if they're coach of two?

eldering: Indeed. I'm not sure what my choice personally would be if a coach is associated to two teams, but I think we should give the API implementer and/or admin some flexibility, and not make it a hard requirement that a coach MUST have a non-empty `team_id` field.
But I admit that I might be nitpicking a bit too much here...

niemela: I agree with @eldering (but I also agree that this might be too much nitpicking, and I would be fine with merging as is)

clevengr: In case people wonder, I'll point out that it is quite common for a coach to be the coach of more than one team, even at the World Finals.  Coach Abd Wahad from the ACPC region is an example; he regularly brings multiple teams (from different universities) to WF.

Aren't we in essence saying that the `team_id` should not be a _field_ but rather should be an _array_ (that is, a list of team ids)?

niemela: > Aren't we in essence saying that the `team_id` should not be a _field_ but rather should be an _array_ (that is, a list of team ids)?

Yes, but that would be a breaking change. *This* PR is a bug fix to the `2022-07` version of the spec (as can be seen by the fact that this is a PR to the `2022-07` branch) . Note that it currently is "Required if role is `team`", but `team` is not one of the allowed values for role, while `contestant` and `coach` are.

The update you imply will come as a separate PR to the `master` branch where we work on the next future version of the spec.

niemela, line 1274:
```suggestion
| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `contestant`.
```

niemela (changes requested):

eldering (approved):

niemela (approved):

PR #122, opened 2022-10-03
Add run_time to example since it is required.

nickygerritsen:
As found by @vmcj


niemela (approved):

niemela:
This should also be added to 2022-07.

deboer-tim (approved):

eldering (approved):

PR #125, opened 2022-10-06
Clarify required endpoints

deboer-tim:
- Removed TODO (still in master branch).
- Removed "The following list of API endpoints should be supported", because that seems wrong, or confusing at best.
- Moved 'singular noun' sentence down because it seemed out of place, and added missing `account`.
- Reworded the metadata endpoint sentence.
- Added section on 'Required and Optional API'. There is some redundancy with the metadata paragraph, but I felt it was appropriate.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -507,7 +507,7 @@ The endpoints can be categorized into 4 groups as follows:
     awards, commentary;
   - Aggregate data: scoreboard, event-feed.
 
-Metadata is data about the API, and are the only required API endpoints. 
+The metadata endpoints contain data about the API, and are the only required API endpoints. 
 They are not included in the event feed. The access endpoint specifies
 which other endpoints are offered by the API.
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -531,7 +531,7 @@ NDJSON format.
 Note that `api`, `access`, `account`, `state`, `scoreboard`, and `event-feed`
 are singular nouns and indeed contain only a single object.
 
-### Required and Optional API
+### Required and optional endpoints
 
 The only required endpoints are metadata: `api` and `access`. The
 only required property is `id` (if it exists on the endpoint).
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -331,8 +331,6 @@ access to.
 | [admin_submit](#modifying-submissions)    | Submit as an admin                           |
 | [admin_clar](#modifying-clarifications)   | Submit clarifications as an admin            |
 
-TODO - add capabilities related to team view, awards, and freeze time.
-
 ### Notification format
 
 There are two mechanisms that clients can use to receive notifications
@@ -495,10 +493,6 @@ are meant to ease extensibility:
 
 ## Interface specification
 
-The following list of API endpoints should be supported. Note that `access`,
-`state`, `scoreboard` and `event-feed` are singular nouns and indeed
-contain only a single object.
-
 All endpoints should support `GET`; specific details on other methods
 are mentioned below.
 
@@ -513,9 +507,9 @@ The endpoints can be categorized into 4 groups as follows:
     awards, commentary;
   - Aggregate data: scoreboard, event-feed.
 
-Metadata is data about the API. These are not included in the event feed and
-are always required to be available. The access endpoint specifies which other
-endpoints are offered by the API. 
+Metadata is data about the API, and these are the only required API endpoints. 
+These are not included in the event feed. The access endpoint specifies
+which other endpoints are offered by the API.
 
 Configuration is normally set before contest start. Is not expected to,
 but could occasionally be updated during a contest. It does not have
@@ -533,6 +527,33 @@ Aggregate data: Only `GET` makes sense. These are not included in the
 event feed, also note that these should not be considered proper REST
 endpoints, and that the `event-feed` endpoint is a streaming feed in
 NDJSON format.
+
+Note that `access`, `account`, `state`, `scoreboard` and `event-feed`
+are singular nouns and indeed contain only a single object.
+
+### Required and Optional API
+
+The only required endpoints are metadata: `api` and `access`. The
+only required property is `id` (if it exists on the endpoint).
+
+All other endpoints and properties are optional. The only exceptions
+to this are obvious logical requirements (for example, a submission
+must come from a team) and the requirement for referential integrity
+(for example, if a submission has a team_id, then teams must be supported).
+`access` exists so that you can discover which endpoints and properties
+are supported by a given provider.
+
+In practice there are different types of providers that will offer
+similar sets of endpoints. Some examples:
+ - A contest management system will support at least contests and
+   teams, and may support other configuration endpoints.
+ - A Contest Control System (CCS) will support at least submissions,
+   judgements, and dependencies of these, as well as scoreboard and
+   event-feed.
+
+Separate specifications (for example, the CCS System Requirements)
+will provide more information on which endpoints and properties
+can be expected, often in the form of a minimal `access` response.
 
 ### Table column description
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -536,7 +536,7 @@ are singular nouns and indeed contain only a single object.
 
 The only required endpoints are metadata: `api` and `access`.
 The only requirements for properties are that collections must have
-an `id` property, and all endpoints that are supported must
+an `id` property.
 have at least some property, i.e. an endpoint can't be empty.
 [Referential integrity](#referential-integrity) must also be kept
 (for example, if a submission has a team_id, then teams must be supported).
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -7,7 +7,7 @@ permalink: /contest_api
 ## Introduction
 
 This page describes an API for accessing information provided by a
-[Contest Control System](ccs_system_requirements) or
+[Contest Control System](ccs_system_requirements) (CCS) or
 [Contest Data Server](https://tools.icpc.global/cds/).
 Such an API can be used by a multitude of clients:
 
@@ -493,7 +493,7 @@ are meant to ease extensibility:
 
 ## Interface specification
 
-The following list of API endpoints may be supported.
+The following list of API endpoints may be supported, as detailed below.
 All endpoints should support `GET`; specific details on other methods
 are mentioned below.
 
@@ -549,9 +549,9 @@ In practice there are different types of providers that will offer
 similar sets of endpoints. Some examples:
  - A contest management system will support at least contests and
    teams, and may support other configuration endpoints.
- - A Contest Control System (CCS) will support at least submissions,
-   judgements, and dependencies of these, as well as scoreboard and
-   event-feed.
+ - A CCS will support at least submissions, judgements, and
+   dependencies of these. It will likely support a scoreboard, and
+   usually an event-feed.
 
 Separate specifications (for example, the CCS System Requirements)
 will provide more information on which endpoints and properties
@@ -575,7 +575,7 @@ Note that all results returned from endpoints:
   - Must only have `null` values if the type of the property is `<type> ?`.
   - Must contain all properties specified in the [Access](#access) endpoint
     that have non-`null` values. 
-  - Should not contain any properties not specified in the [Access](#access) 
+  - Must not contain any properties not specified in the [Access](#access) 
     endpoint.
 
 ### Filtering
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -537,7 +537,6 @@ are singular nouns and indeed contain only a single object.
 The only required endpoints are metadata: `api` and `access`.
 The only requirements for properties are that collections must have
 an `id` property.
-have at least some property, i.e. an endpoint can't be empty.
 [Referential integrity](#referential-integrity) must also be kept
 (for example, if a submission has a team_id, then teams must be supported).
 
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -534,6 +528,33 @@ event feed, also note that these should not be considered proper REST
 endpoints, and that the `event-feed` endpoint is a streaming feed in
 NDJSON format.
 
+Note that `access`, `account`, `state`, `scoreboard` and `event-feed`
+are singular nouns and indeed contain only a single object.
+
+### Required and Optional API
+
+The only required endpoints are metadata: `api` and `access`. The
+only required property is `id` (if it exists on the endpoint).
```
niemela, line 537: The claim about properties is technically true, but I find it confusing. I would like to improve or remove that sentence.

Contest_API.md
```diff
@@ -534,6 +528,33 @@ event feed, also note that these should not be considered proper REST
...
+
+All other endpoints and properties are optional. The only exceptions
+to this are obvious logical requirements (for example, a submission
+must come from a team) and the requirement for referential integrity
```
niemela, line 541: "obvious logical requirements" seems very undefined. 

I'm not even sure I agree that the example is *absolutely* necessary.

niemela (changes requested):

Contest_API.md
```diff
@@ -513,9 +507,9 @@ The endpoints can be categorized into 4 groups as follows:
     awards, commentary;
   - Aggregate data: scoreboard, event-feed.
 
-Metadata is data about the API. These are not included in the event feed and
-are always required to be available. The access endpoint specifies which other
-endpoints are offered by the API. 
+Metadata is data about the API, and these are the only required API endpoints. 
+These are not included in the event feed. The access endpoint specifies
```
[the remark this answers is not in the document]
nickygerritsen: @johnbrvc actually asked me this exact question in a Slack DM, he would like to have api info at the beginning of the feed. I told him to put it on the next agenda.

Contest_API.md
```diff
@@ -534,6 +528,33 @@ event feed, also note that these should not be considered proper REST
 endpoints, and that the `event-feed` endpoint is a streaming feed in
 NDJSON format.
 
+Note that `access`, `account`, `state`, `scoreboard` and `event-feed`
+are singular nouns and indeed contain only a single object.
```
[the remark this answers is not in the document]
nickygerritsen: Up to this point we never used `api` in that sense. And is it `api` or `api-information`? Since the section is called "API information", not "API"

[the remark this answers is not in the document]
niemela: It's `api` in the CPF, so I would go with that.

Contest_API.md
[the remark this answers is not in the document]
deboer-tim: I added it to the agenda, but I am assuming that's a 'future draft' topic so resolving this. Accepting commit.

Contest_API.md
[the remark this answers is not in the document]
deboer-tim: Accepted comment that added it.

Contest_API.md
deboer-tim: How? I couldn't get past the fact that it is indeed true.

Contest_API.md
deboer-tim: I agree, but I couldn't come up with a better term that would be any more specific or meaningful. This is really just a weak "it would be silly to do X". Do we just remove this entirely and the only requirements are ids and referential integrity?

Contest_API.md
```diff
@@ -495,10 +493,6 @@ are meant to ease extensibility:
 
 ## Interface specification
 
-The following list of API endpoints should be supported. Note that `access`,
```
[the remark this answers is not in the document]
niemela: Is "may" better? Because "should" is "wrong", and "could" or "may" isn't. (I agree that "could" feels strange).

Contest_API.md
niemela: Fair point :).

Could we maybe say something like "The only requirements for properties is that collections must have the `id` property", and maybe also "all endpoints that are supported must have at least some property, i.e. an endpoint can't be empty."

Contest_API.md
niemela: I would suggest that we just remove this entirely.

Contest_API.md
```diff
@@ -513,9 +507,9 @@ The endpoints can be categorized into 4 groups as follows:
     awards, commentary;
   - Aggregate data: scoreboard, event-feed.
 
-Metadata is data about the API. These are not included in the event feed and
-are always required to be available. The access endpoint specifies which other
-endpoints are offered by the API. 
+Metadata is data about the API, and are the only required API endpoints.
```
eldering, line 510:
```suggestion
The metadata endpoints contain data about the API, and are the only required API endpoints. 
```

Contest_API.md
```diff
@@ -513,9 +507,9 @@ The endpoints can be categorized into 4 groups as follows:
     awards, commentary;
   - Aggregate data: scoreboard, event-feed.
 
-Metadata is data about the API. These are not included in the event feed and
-are always required to be available. The access endpoint specifies which other
-endpoints are offered by the API. 
+Metadata is data about the API, and are the only required API endpoints. 
+They are not included in the event feed. The access endpoint specifies
+which other endpoints are offered by the API.
```
eldering, line 512: I'd like to specify here or further down that anything specified in `access` must be provided by this API service. For example:
```suggestion
which other endpoints are offered by the API. That is, any endpoints and their properties listed in `access` must be provided (possibly with a `null` value when the property is optional), and only these endpoints and properties. 
```
I'd like this to be rigid in this sense, so that a client knows exactly what to expect, but throwing it out there for discussion.

Contest_API.md
eldering: I'm fine with removing this, but also see my remark above about relating data in `access` to requiring those endpoints and properties.

Contest_API.md
deboer-tim: Updated to this. It felt a little odd to have the referential integrity lower down, so I moved it to this paragraph as well. Please review.

Contest_API.md
deboer-tim: Removed entirely while responding to comment above. /access suggestion was accepted. Please re-review.

Contest_API.md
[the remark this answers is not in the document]
eldering: Maybe add:
`may be supported, as detailed below`?

Contest_API.md
```diff
@@ -513,9 +508,9 @@ The endpoints can be categorized into 4 groups as follows:
     awards, commentary;
   - Aggregate data: scoreboard, event-feed.
 
-Metadata is data about the API. These are not included in the event feed and
-are always required to be available. The access endpoint specifies which other
-endpoints are offered by the API. 
+The metadata endpoints contain data about the API, and are the only required API endpoints. 
+They are not included in the event feed. The access endpoint specifies
+which other endpoints are offered by the API. That is, any endpoints and their properties listed in `access` must be provided (possibly with a `null` value when the property is optional), and only these endpoints and properties.
```
eldering, line 513: When rereading the spec, I now see that the requirement on properties is already specified under `### Table column description` below, and slightly differently, because it says that the properties in `access` *must* be provided, and any that are not there *should* not.
I don't think it's bad to have this repeated, but I think we should put them in sync. I'd be in favor of making it a *must* to have `access` exactly in sync with which properties are there, so only properties listed in `access` may be provided.

deboer-tim: I definitely agree these places should match, but I'm struggling with the _must_. Compared with other places, having an extra/unexpected attribute doesn't seem like a particularly harmful thing. If you're the kind of client that called `access`, you can easily/safely ignore it if you want to. Yes, you shouldn't do that, but it doesn't really break anything.

@niemela I checked the git blame :) and you wrote the 'should' originally. How do you or @nickygerritsen feel?

nickygerritsen: I’d say I agree: if you can only put fields that are in access that makes extensibility hard right? Say domjudge has a field “name” on account that isn’t in the spec, I don’t necessarily want to put it in access. But then if we add that name field to the spec (which we did right?) are we illegal?

eldering: I'd say: if you provide an endpoint/property that's outside the spec, then you should still put it in `access`, and then I don't see the problem of making it a *must* to put all your endpoints and properties in `access`. Is there a reason leaving anything that's not defined in the spec out of `access`?

eldering: So to me `access` is a contract between the server and client about what the server actually provides.

deboer-tim: I agree all properties (even ones that are custom/beyond the spec) _should_ go in `access`. This lets clients know they exist and may see them, even if they don't know how to handle or will ignore them. I think we should do a separate PR to make this explicit.

My point was that a CCS not putting account name in `access` (or maybe just 'not putting it in _yet_') shouldn't cause harm to any client, so I wasn't convinced it should be a 'must'. To me that property is not spec-compliant, but I would still say the CCS is.

niemela: I agree with @eldering .

I don't see the benefit of not making this a 'must'.  The spec is actually *very* lenient in what is allowed. Almost everything that's listed is optional, and you may add literally anything else. Requiring that you declare what you have chosen to support from the spec and add to the spec seems like a very reasonable requirement.

Contest_API.md
```diff
@@ -534,6 +529,34 @@ event feed, also note that these should not be considered proper REST
...
 
+Note that `api`, `access`, `account`, `state`, `scoreboard`, and `event-feed`
+are singular nouns and indeed contain only a single object.
+
+### Required and optional endpoints
+
+The only required endpoints are metadata: `api` and `access`.
+The only requirements for properties are that collections must have
+an `id` property, and all endpoints that are supported must
+have at least some property, i.e. an endpoint can't be empty.
```
niemela, line 540: Is it clear (enough) that this property could be the `id`? I.e. for collections the "and all endpoints that are supported must have at least some property" requirement is actually a no-op?

Contest_API.md
```diff
@@ -534,6 +529,34 @@ event feed, also note that these should not be considered proper REST
...
+`access` exists so that you can discover which endpoints and properties
+are supported by a given provider.
+
+In practice there are different types of providers that will offer
+similar sets of endpoints. Some examples:
+ - A contest management system will support at least contests and
+   teams, and may support other configuration endpoints.
+ - A Contest Control System (CCS) will support at least submissions,
+   judgements, and dependencies of these, as well as scoreboard and
+   event-feed.
```
niemela, line 554: That's not at all certain.  You can nitpick some unreasonable example where pretty much all of these examples could be theoretically broken, but a CCS could *absolutely* not support event-feed, and maybe even scoreboard.

niemela (changes requested):

Contest_API.md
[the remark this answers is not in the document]
deboer-tim: Added in new commit.

Contest_API.md
deboer-tim: Ok, other section was updated so that it's a must in both instances.

Contest_API.md
deboer-tim: I agree with what you're saying, but we have these in practise and these are only examples. 
I updated it so that scoreboard is 'likely' and event-feed is 'usual'. Is that downplayed enough?

Contest_API.md
deboer-tim: How would you word it? I'm not sure there will be cases where only an id will be useful, so I kind of felt like it was ok if this was technically valid but not obvious that was the case :). Or that we should even enforce one property in addition to the id.

eldering (approved):
LGTM

Contest_API.md
niemela: If anything I would lean the other way, i.e. drop ", and all endpoints that are supported must
have at least some property, i.e. an endpoint can't be empty." 

Having an empty endpoint is useless, but it doesn't break anything. (Also, there is a trivial, and similarly useless, workaround. You could add the "ignoreme" property to an otherwise empty endpoint.).

Contest_API.md
```diff
@@ -534,6 +529,34 @@ event feed, also note that these should not be considered proper REST
...
 NDJSON format.
 
+Note that `api`, `access`, `account`, `state`, `scoreboard`, and `event-feed`
+are singular nouns and indeed contain only a single object.
+
+### Required and optional endpoints
+
+The only required endpoints are metadata: `api` and `access`.
+The only requirements for properties are that collections must have
+an `id` property, and all endpoints that are supported must
```
niemela, line 539:
```suggestion
an `id` property.
```

Contest_API.md
niemela, line 540:
```suggestion
```

Contest_API.md
niemela: Sure.

niemela (changes requested):
Remove the requirement for non-empty endpoints.

Contest_API.md
deboer-tim: Ok, fair enough. Suggested changes committed.

niemela (approved):

Contest_API.md
nickygerritsen, line 513: Can we make this line less long, i.e. add some enters?

nickygerritsen (approved):

PR #126, opened 2022-10-06
Allow person to be on multiple teams

deboer-tim:
Same change as in 2022_07 clarification, but make team_ids an array, and make it required for coaches.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1271,7 +1271,7 @@ Properties of person objects:
 | :---------- | :-------------- | :----------
 | id          | ID              | Identifier of the person.
 | icpc\_id    | string ?        | External identifier from ICPC CMS.
-| team\_ids   | array of ID ?   | [Team](#teams) of this person. Required iff role is `contestant` or `coach`.
+| team\_ids   | array of ID ?   | [Team](#teams) of this person. Required to be non-empty iff role is `contestant` or `coach`.
 | name        | string          | Name of the person.
 | title       | string ?        | Title of the person, e.g. "Technical director".
 | email       | string ?        | Email of the person.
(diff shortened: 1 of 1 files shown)


niemela (dismissed):

nickygerritsen (approved):

Contest_API.md
```diff
@@ -1271,12 +1271,12 @@ Properties of person objects:
 | :---------- | :-------------- | :----------
 | id          | ID              | Identifier of the person.
 | icpc\_id    | string ?        | External identifier from ICPC CMS.
-| team\_id    | ID ?            | [Team](#teams) of this person. Required iff role is `team`.
+| team\_ids   | array of ID ?   | [Team](#teams) of this person. Required iff role is `contestant` or `coach`.
```
nickygerritsen, line 1274: Maybe
```suggestion
| team\_ids   | array of ID ?   | [Team](#teams) of this person. Required to be non-empty iff role is `contestant` or `coach`.
```
? Technically an empty list is also a value but that is not what we want.

nickygerritsen (dismissed):

Contest_API.md
niemela: I think we should allow empty lists. (we do for all other `array of ID`).

That said, the fact that it's nullable *and* allowed to be empty means that there are two ways to specify the same thing. Maybe it should not be nullable? (If we do that change we should do the same change in all the similar places)

nickygerritsen: I was talking about the contestant or role persons. I think we wanted to say that those NEED a team.

niemela: Ah, yes, of course. I agree.

deboer-tim:
Accepted 'non-empty' commit, please reapprove.

nickygerritsen (approved):

niemela (approved):

PR #133, opened 2023-01-23
Clarify and reformulate requirements of href in API/package.

eldering:


nickygerritsen (approved):

deboer-tim (approved):

nickygerritsen (approved):

Contest_API.md
```diff
@@ -253,7 +253,7 @@ Properties for file reference objects:
 
 | Name     | Type      | Description
 | -------- | --------- | -----------
-| href     | string ?  | URL where the resource can be found. Relative URLs are relative to the `baseurl`. Must point to a file of intended mime-type. Resource must be accessible using the exact same (possibly none) authentication as the call that returned this data.
+| href     | string ?  | URL where the resource can be found. Relative URLs are relative to the `baseurl`. Must point to a file of intended mime-type. Resource must be accessible using the exact same (possibly none) authentication as the call that returned this data. Required when served from an API, but optional inside a [Contest Package].
```
niemela, line 256: I don't think this should be added here. IMO it's more confusing.

Contest_Package.md
```diff
@@ -40,37 +40,35 @@ Information in the API is always either in JSON format,
...
 
-
-Files referenced in `api.json` and `contest.json` are stored as `api/<filename>`
-and `contest/<filename>` respectively, and files referenced in
-`<endpoint>.json` are stored as `<endpoint>/<id>/<filename>`, where:
-
-- `<id>` is the ID of the endpoint object the reference is in.
-- `<filename>` is the filename specified in the file reference object.
-
 Some of these API endpoints are often written by humans in a package. For this
```
niemela, line 43: I realize this was not changed in the PR, but "these API endpoints are often written by humans in a package" makes no sense?

Contest_Package.md
```diff
@@ -40,37 +40,35 @@ Information in the API is always either in JSON format,
...
-Optionally one could create a Shallow Package by not storing the files, in which
-case the URLs must be valid. This could be useful in some cases where the size
-of the package matters.
+The file reference may optionally contain a URL in the `href`
+property, for example when the package is archived from a Contest API.
+However, if a file is present in the package at the expected storage
+location, then that should be used and the original URL ignored. If no
+such file exists, then the original URL should be used to retrieve the
+file. This option could be used to create a shallow package, when the
+size of the package matters.
```
niemela, line 71: I think this is *less* clear than the old version.

niemela (changes requested):

Contest_Package.md
eldering: I've rewritten it. Hopefully it's clearer now.

Contest_Package.md
eldering: Rewritten, keeping most of the old text to explain the why.

Contest_API.md
eldering: I've kept this for now as I think we *do* need to mention this somewhere and we don't have a separate document to specify the requirements for using the "bare specification" in the context of serving it as an API.

Contest_Package.md
niemela: I think it's clearer now.

Contest_Package.md
niemela: I think this change makes you more happy than it makes me unhappy, so in total happiness increases. I'm fine with this.

Contest_API.md
niemela: I still don't like this. I suggest the following (inspired by @nickygerritsen)

- Make `href` non-optional
- Put the added sentence in parathesis with a "Note: ".

```suggestion
| href     | string    | URL where the resource can be found. Relative URLs are relative to the `baseurl`. Must point to a file of intended mime-type. Resource must be accessible using the exact same (possibly none) authentication as the call that returned this data. (Note: this is actuallu optional inside a [Contest Package]).
```

nickygerritsen (approved):

PR #134, opened 2023-01-24
Add capability to thaw a contest.

nickygerritsen:


Contest_API.md
```diff
@@ -760,6 +762,18 @@ can only be non-null when start time is null.
 The request should fail with a 403 error code if the contest is started or within 30s of
 starting, or if the new start time is in the past or within 30s.
 
+Clients with the `contest_thaw` [capability](#capabilities) have the ability to
+set a time when the contest will be thawed via a PATCH method.
+
+The Patch must include a valid JSON object with exactly three properties:
+the contest `id` (used for verification) and a `contest_thaw_time`, which can
+either be a `<TIME>` value or the specialstring value `"now"`, which indicates
```
deboer-tim, line 770: Typo: special string

I'm not sure I love 'now' vs setting the current time, but I'm ok with it. FWIW the CDS has similar extension for start time where you can essentially say "start in 5m" or "reduce the pause time by 30s" to make clients life easier, and I waffle between thinking 'it should be removed' and 'it makes things so much easier for clients/users'.

Contest_API.md
```diff
@@ -760,6 +762,18 @@ can only be non-null when start time is null.
...
+the contest should be thawed at the current time.
+
+The request should fail with a 4xx error code if the contest can't be thawed at the given
```
deboer-tim, line 773: Shouldn't this be 403 like the start time?

Contest_API.md
nickygerritsen: I’m struggling with this as well. But I van imagine not all CCSes can start at any time, and can only do now. If we think we all can do any time I can drop it

Contest_API.md
nickygerritsen: I swear I saw 4xx but I was just blind.

Contest_API.md
deboer-tim: Fair enough. If more CCSs are likely to implement 'now' but not timed then let's go with that.

Contest_API.md
deboer-tim: I've tried that explanation but it works better when you do it. 😂

Contest_API.md
nickygerritsen: Yeah. DOMjudge can do 'any' time, so if the other big CCS'es can also do it, I'm fine dropping `now`.

Contest_API.md
```diff
@@ -760,6 +762,18 @@ can only be non-null when start time is null.
 The request should fail with a 403 error code if the contest is started or within 30s of
 starting, or if the new start time is in the past or within 30s.
 
+Clients with the `contest_thaw` [capability](#capabilities) have the ability to
+set a time when the contest will be thawed via a PATCH method.
+
+The Patch must include a valid JSON object with exactly three properties:
```
eldering, line 768: Typo:
```suggestion
The Patch must include a valid JSON object with exactly two properties:
```

Contest_API.md
eldering: Hmmm, but what should happen if the client sends a thaw time that is a few seconds in the past, e.g. due to network delay? In principle, I think that this is (or could be considered) invalid, because the CCS may not be able to retroactive thaw the contest at a past time. So from that point of view `now` seems a good option...

nickygerritsen: I kinda left that in the middle on purpose by saying that an invalid time should give 403. But I do agree thawing in the past should not be possible. So yeah, maybe you are right and we should keep `now` if you want to thaw... now.

meisterT (approved):

deboer-tim (dismissed):
Approved. I was just trying to figure out the correct order of events though. I assume in a simple contest a client would trigger the thaw, the server sends the thaw state event, followed by all the judgements from during the freeze, then the end of contest.

If a resolver is directly connected to the CCS, we can't trigger the thaw first b/c then other clients would see all judgements, so I guess it would trigger thaw at the end?

If a resolver is connected via the CDS, it could likely send a thaw event, let the judgements out slowly (on clicks) to all clients, then end of contest at the end. If other systems want to do this we could spec a separate 'expose judgement' capability.

Contest_API.md
```diff
@@ -760,6 +762,23 @@ can only be non-null when start time is null.
...
+The request should succeed with a 204 response code with no body if the server changed the
+thaw time to the time specified.
+
+The server may also thaw the contest at the current server time if the provided `contest_thaw_time`
+is in the past. In that case the server must reply with a 200 response code and the modified contest
+as body, so the client knows the server used a different thaw time.
+
+The request should fail with a 403 error code if the contest can't be thawed at the given
+time, for example because the thaw time is before the contest end, the contest is already thawed
+or the server does not support setting a specific thaw time.
```
eldering, line 780: Nitpick to make it a bit clearer (I think):
```suggestion
or the server does not support setting this specific thaw time.
```

eldering (approved):

eldering:
> Approved. I was just trying to figure out the correct order of events though. I assume in a simple contest a client would trigger the thaw, the server sends the thaw state event, followed by all the judgements from during the freeze, then the end of contest.

I assume instead of "end of contest" you mean `end_of_updates` (also below)?

> If a resolver is directly connected to the CCS, we can't trigger the thaw first b/c then other clients would see all judgements, so I guess it would trigger thaw at the end?

Yes, I guess so, and the resolver would have to be connected with jury/admin privileges.

> If a resolver is connected via the CDS, it could likely send a thaw event, let the judgements out slowly (on clicks) to all clients, then end of contest at the end. If other systems want to do this we could spec a separate 'expose judgement' capability.

Sounds reasonable to me. There's nothing that forces the CDS to expose all judgements immediately. I'm not sure I'd immediately add a separate capability for this in the spec, but of course feel free to add it to the CDS.

deboer-tim:
Yes and agreed 👍🏼

eldering (dismissed):

niemela (approved):

eldering (approved):

PR #143, opened 2023-04-05
Add label to team, making the ID be 'something else'.

nickygerritsen:
@niemela I know you were tasked to do this but maybe forgot (or didn't have time yet), but we want to test this at NAC, so we first need the spec to have it.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1268,14 +1268,15 @@ The following endpoints are associated with teams:
 | Endpoint                     | Mime-type        | Description
 | :--------------------------- | :--------------- | :----------
 | `/contests/<id>/teams`       | application/json | JSON array of all teams with properties as specified by `/access`.
-| `/contests/<id>/teams/<id>`   | application/json | JSON object representing a single team with properties as specified by `/access`.
+| `/contests/<id>/teams/<id>`  | application/json | JSON object representing a single team with properties as specified by `/access`.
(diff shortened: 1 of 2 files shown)


deboer-tim (dismissed):
Approved, but shouldn't it be non-optional?

nickygerritsen:
> Approved, but shouldn't it be non-optional?

I was wondering the same. I guess this is a field that if you have it, you better have it for all teams right? So then we should make it non optional?

deboer-tim:
> > Approved, but shouldn't it be non-optional?
> 
> I was wondering the same. I guess this is a field that if you have it, you better have it for all teams right? So then we should make it non optional?

Yes, that's what I was assuming.

niemela (approved):

meisterT:
It's hard for me to understand when to use what of id, label, name and display_name. What gap are we filling with this PR?

nickygerritsen:
The gap we are filling is that for *all other types* we get from the CMS (groups, organizations) we use the ICPC ID as `ID`, which never changes (or shouldn't) but for the teams we use the 'seat number', but we normally don't have that unless we are at the WF. The `label` is very similar (and hence named the same way) as the `label` on a problem.
Basically: ID is internal and no one should really know it (just like problem ID, group ID, etc.) and label is what teams know themselves.

PR #145, opened 2023-04-11
Judge capabilities

deboer-tim:
Edited:
team_clar has been changed to a generic post_clar, and a post_comment capability with access equivalent to what a judge/analyst should be able to do has been added, and described similar to other capabilities.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -329,7 +329,8 @@ access to.
 | [contest\_start](#modifying-contests)      | Control the contest's start time
 | [contest\_thaw](#modifying-contests)       | Control the contest's thaw time
 | [team\_submit](#modifying-submissions)     | Submit as a team
-| [team\_clar](#modifying-clarifications)    | Submit clarifications as a team
+| [post\_clar](#modifying-clarifications)    | Submit clarifications
+| [post\_comment](#modifying-clarifications) | Submit commentary
 | [proxy\_submit](#modifying-submissions)    | Submit as a shared team proxy
 | [proxy\_clar](#modifying-clarifications)   | Submit clarifications as a shared team proxy
 | [admin\_submit](#modifying-submissions)    | Submit as an admin
@@ -1540,7 +1541,7 @@ zip archive. These must be stored directly from the root of the zip
 file, i.e. there must not be extra directories (or files) added unless
 these are explicitly part of the submission content.
 
-#### Modifying submissions 
+#### Modifying submissions
 
 To add a submission, clients can use the `POST` method on the submissions endpoint or the
 `PUT` method directly on an object url. One of the following [capabilities](#capabilities)
@@ -1816,7 +1817,7 @@ is required to add clarifications, with descriptions below:
 
 | Name               | Description
 | :----------------- | :----------
-| team\_clar         | POST a clarification as a team
+| post\_clar         | POST a clarification
 | proxy\_clar        | POST a clarification as a proxy (able to submit on behalf of team(s))
 | admin\_clar        | POST or PUT a clarification as an admin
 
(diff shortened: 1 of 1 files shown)


nickygerritsen (approved):

Contest_API.md
```diff
@@ -330,6 +330,8 @@ access to.
 | [contest\_thaw](#modifying-contests)       | Control the contest's thaw time
 | [team\_submit](#modifying-submissions)     | Submit as a team
 | [team\_clar](#modifying-clarifications)    | Submit clarifications as a team
+| [judge\_clar](#modifying-clarifications)   | Submit clarifications as a judge
```
eldering, line 333: I think we should merge the `team_clar` and `judge_clar` capabilities into one. I don't really see why you'd want to support one but not the other and they're not really different in complexity, except maybe that a judge may respond to an existing clarification and a team might not(?) This is for example different for the `submit_*` capabilities, where the admin or proxy one can override more attributes and/or have slightly different semantics.

Contest_API.md
```diff
@@ -330,6 +330,8 @@ access to.
...
+| [judge\_comment](#modifying-commentary)    | Submit commentary as a judge
```
eldering, line 334: Does it make sense to rename this to `post_comment` or so? Commenting would typically be reserved for judges, and the capability can anyways be role restricted, but I don't think that the semantics or details of this capability would be different if another role (say public) would have it.

Contest_API.md
```diff
@@ -2096,6 +2102,33 @@ tags below are used.
...
+with description below:
+
+| Name               | Description
+| :----------------- | :----------
+| judge\_comment     | POST a commentary as a judge
+
+All requests must include a valid JSON object with the same properties as the commentary
+endpoint returns from a `GET` request with the following exceptions:
+
+- The `judge\_comment` capability only has access to `POST`. `id`, `time`,
```
eldering, line 2118: Nitpick: since no `PUT` is defined, this is not really necessary. But I'm fine with leaving it in, for in case we add that.

Contest_API.md
deboer-tim: So one capability, but it says if your account type is team you can do X and if your account is judge you have slightly different access? Yes, I can do that, just wasn't sure that everyone would be willing to implement both at the same time - but I guess you could implement one and only advertise the capability via /access to that account type.

Contest_API.md
deboer-tim: As above, agreed.

Contest_API.md
deboer-tim: This was copied from similar capabilities, will try to improve as part of a rewrite.

Contest_API.md
deboer-tim: PR updated, please re-review.

Contest_API.md
deboer-tim: PR updated, please re-review.

Contest_API.md
```diff
@@ -1820,10 +1821,13 @@ exceptions:
 - When a property value would be null it is optional - you do not need to
   include it. e.g. if a clarification is not related to a problem you can
   choose to include or exclude the `problem_id`.
-- The `team_clar` capability only has access to `POST`. `id`, `to_team_id`,
-  `time`, and `contest_time` must not be provided. `from_team_id` may be
+- The `post_clar` capability only has access to `POST`. `id`, ,
```
eldering, line 1824: Remove left-over comma:
```suggestion
- The `post_clar` capability only has access to `POST`. `id`,
```

Contest_API.md
```diff
@@ -2108,6 +2112,33 @@ tags below are used.
...
+with description below:
+
+| Name               | Description
+| :----------------- | :----------
+| post\_comment      | POST a commentary message
+
+All requests must include a valid JSON object with the same properties as the commentary
+endpoint returns from a `GET` request with the following exceptions:
+
+- The `post\_comment` capability only has access to `POST`. `id`, `time`,
```
eldering, line 2128: I think this first sentence should be removed. In the clarifications endpoint it made sense since there was also a `{proxy,admin}_clar` capability with PUT access, but here there is only POST.
```suggestion
- `id`, `time`,
```

niemela:
@deboer-tim What is the status of this? Can we close?

deboer-tim:
Discuss on the next call? This is implemented on the CDS and is required API support for clients to submit commentary - but I'm not aware of anyone else looking at / implementing this so not sure if it's worth spec or should just be a CDS-specific extension for now.

deboer-tim:
Fixed remaining comment, and rebased on latest. Moving to review status.

Contest_API.md

Contest_API.md

Contest_API.md
```diff
@@ -2115,6 +2119,32 @@ tags below are used.
 | accepted-winner         | A submission was accepted that changed the set of teams currently in the lead.
 | accepted-\<award>       | A submission was accepted that changed the set of teams awarded \<award>. Note that the above 4 are special cases of this.
 
+#### Modifying commentary
```
eldering, line 2122: Since this only about posting new commentary, not editing existing ones, I'd suggest to name it:
```suggestion
#### Adding commentary 
```
or something similar.

deboer-tim: As discussed, won't change since this title matches other sections.

niemela (approved):

Contest_API.md
eldering: As discussed, all other endpoints use the same name, so drop this.

eldering (approved):

Issue #150, opened 2023-05-14
Unsupported ICPC rules modification 


kunyavskiy:
Some contest control systems do not round the penalty for each problem to the minute but use some other penalty resolution. 

The main option I have seen is to sum submission times with seconds resolution and round down to the minute at the end, but just using penalty as seconds also makes sense.

Also, some contests (e.g. Atcoder rounds) use last-submission time instead of sum-of-submissions. 

I think it should be possible to specify these changes in the Contest object, a `penlty_time` is specified, not forced to be equal to 20 minutes. 

So my proposal is, to add the following fields in the contest object:
1. `problem_penalty_resolution` it can be either an integer (in seconds or in milliseconds?) or enum (MINUTE, SECOND, etc). 
2. `total_penalty_resolution` in a similar way
3. `penalty_algorithm` which is enum, whether `sum` or `last`, maybe some more options if someone comes up with them. 

This change is easy to implement in CCS'es - they can just send objects correctly to them not making it configurable, if it was not. And reasonably easy to implement in scoreboard calculation.  

Questions need to be accurate with:
We should decide if `penalty_time` is specified in minutes or in `problem_penalty_resolution` units. Both make sense to me.

niemela:
> Questions need to be accurate with:
> We should decide if `penalty_time` is specified in minutes or in `problem_penalty_resolution` units. Both make sense to me.

Since #151, `penalty_time` is now specified as `RELTIME`. So, neither.

> Also, some contests (e.g. Atcoder rounds) use last-submission time instead of sum-of-submissions.

I thought AtCoder used scoring problems? As such they would not have "penalty" in the sense we talk about here? Could you elaborate on what the exact rules are for AtCoder rounds?

> So my proposal is, to add the following fields in the contest object:
> 
>    1. `problem_penalty_resolution` it can be either an integer (in seconds or in milliseconds?) or enum (MINUTE, SECOND, etc).

Maybe `contest_time_resolution` would be more consistent with current terminology (See [scoring section](https://ccs-specs.icpc.io/draft/ccs_system_requirements#scoring) of CCS requirements)?  An argument can also be made for `penalty_time_resolution`.

Most consistent with the changes in #151 would be to specify it as `RELTIME`. I think that would work.

I think we should require that `penalty_time` is an integer multiple of this. Thoughts on that?

>    2. `total_penalty_resolution` in a similar way

Why would you want this to be different?
 
[...]

PR #151, opened 2023-06-12
Penalty time in RELTIME

deboer-tim:
Contest penalty_time is the only time that's in integer minutes, and one of only 3 places where we even say minutes in the spec. As per issue #150, it also blocks non-integer penalty times.

This changes it to use RELTIME, which improves both issues. It does not attempt to address anything else in issue 150, namely no attempt to add additional scoring algorithms.


nickygerritsen (approved):

eldering (commented):
I think the `score.total_time` property in the scoreboard row objects should also be changed to `RELTIME`.

And while we're at it, I think we should also change the `time` property in the problem sub-objects in scoreboard rows from `integer` to `RELTIME`.

I think then we've removed all explicit references to time in minutes.

nickygerritsen (commented):
Line 2195 should also be RELTIME I think?

deboer-tim:
Yup, missed that. Fixed now.

nickygerritsen (approved):

eldering (dismissed):

nickygerritsen (approved):

niemela (approved):

Issue #164, opened 2023-11-09
List of Capabilities is inconsistent with example


johnbrvc:
The `/contests/contestid/access` endpoint description has an example that shows a capability (`patch_time`) that is not part of the **Capabilities** list, rather, the `contest_start` capability is what should probably be used in the example, eg:

```
{
   "capabilities": ["contest_start"],
   "endpoints": [
     { "type": "contests", "properties": ["id","name","formal_name",...]},
     { "type": "problems", "properties": ["id","label",...]},
     { "type": "submissions", "properties": ["id","language_id","reaction",...]}
     ...
   ]
}
```

deboer-tim:
Just to be clear, any system can support capabilities that are not in the spec (like additional objects, this is one of the expected places for systems to extend or try new things for future spec versions), but I 100% agree that our examples should use capabilities that are in the spec.

niemela:
I think this is fixed in master/draft. We should probably fix it in 2023-06 as well.

PR #165, opened 2023-11-13
Fix Issue #164 - change access endpoint example

johnbrvc:
One of the examples for the `Access `endpoint showed a capability of "**patch_time**".  This capability does not appear in the **Capabilities** list that says "_All capabilities are listed in the table below_".  Clearly, "**patch_time**" is not in the list.  For clarity, it makes sense to change the example to use a documented property.   "**contest_start"** makes the most sense here.
This fixes Issue #164

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -706,7 +706,7 @@ Returned data:
 
 ```json
 {
-   "capabilities": ["patch_time"],
+   "capabilities": ["contest_start"],
    "endpoints": [
      { "type": "contests", "properties": ["id","name","formal_name",...]},
      { "type": "problems", "properties": ["id","label",...]},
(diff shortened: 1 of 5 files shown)


deboer-tim (changes requested):
I would approve the contest_start change, but it looks like this branch also picked up several other commits that aren't listed in the description. If you can clean it up (or they get integrated separately) I will approve.

johnbrvc:
> I would approve the contest_start change, but it looks like this branch also picked up several other commits that aren't listed in the description. If you can clean it up (or they get integrated separately) I will approve.

Ugh.  Yeah, I must have branched off of 2023-06. I'll fix that.  Thanks.

johnbrvc:
Apparently, this issue was already fixed in master (draft).  Hence, I am closing this PR.

Issue #166, opened 2023-11-13
Description of the countdown_pause_time is unclear for "/contests" endpoint


johnbrvc:
The description of the `countdown_pause_time` property states:

`The amount of seconds left when countdown to contest start is paused. At no time may both start_time and countdown_pause_time be non-null.`

This is inconsistent with the expected/typical use (case) described below the property list:

 `start_time should be set to null and countdown_pause_time to the number of seconds left to count down. `

The description and the typical use case should be consistent and clear in the meaning of `countdown_pause_time`.

niemela:
I'm sorry, I don't see the inconsistency? Could you elaborate?

PR #167, opened 2023-11-13
Address issue #166 - Change description and typical use case of countdown_pause_time

johnbrvc:
Hopefully, make the explanation of `countdown_pause_time` clearer.  See Issue #166 for details.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -760,8 +760,7 @@ Properties of contest objects:
 | location                     | LOCATION ?      | Location where the contest is held.
 
 The typical use of `countdown_pause_time` is when a contest director wishes to pause the countdown to the start of a contest.  For example, this may occur because of technical
-issues or to make an announcement.  When the contest countdown is paused, the value of `countdown_pause_time` should be set to the number of seconds remaining in the countdown at the
-time of the pause.
+issues or to make an announcement.  When the contest countdown is paused, the value of `countdown_pause_time` should be set to the expected time remaining before the start of the contest after the pause is lifted.
 In this case, the `start_time` must be set to `null`, thereby, setting an undefined `start_time` (the `start_time` is undefined since
 the length of the pause may be unknown).
 The `countdown_pause_time` may be changed to indicate the approximate delay until the contest starts.
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -749,7 +749,7 @@ Properties of contest objects:
 | name                         | string          | Short display name of the contest.
 | formal\_name                 | string ?        | Full name of the contest. Defaults to value of `name`.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
-| countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
+| countdown\_pause\_time       | RELTIME ?       | The number of seconds left when the countdown to the contest start was paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
 | duration                     | RELTIME         | Length of the contest.
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
 | scoreboard\_thaw\_time       | TIME ?          | The scheduled thaw time of the contest, may be `null` if the thaw time is unknown or not set.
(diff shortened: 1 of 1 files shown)


deboer-tim (approved):
This is clearer, thank you.

eldering (dismissed):

Contest_API.md
```diff
@@ -749,7 +749,7 @@ Properties of contest objects:
 | name                         | string          | Short display name of the contest.
 | formal\_name                 | string ?        | Full name of the contest. Defaults to value of `name`.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
-| countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
+| countdown\_pause\_time       | RELTIME ?       | The number of seconds left when the countdown to the contest start was paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
```
niemela, line 752: But this is not a number, it's a `RELTIME`. The old wording; "amount of seconds" is clearly wrong, but maybe "amount of time" is a better fix than "number of seconds"?

There should be a "if the contest is paused" in there. Maybe something like this?:

```suggestion
| countdown\_pause\_time       | RELTIME ?       | The amount of time left when the countdown to the contest start was paused, if the contest is paused, otherwise `null`.
```

Contest_API.md
```diff
@@ -759,11 +759,11 @@ Properties of contest objects:
...
-The expected/typical use of `countdown_pause_time` is that once a
-`start_time` is defined and close, the countdown may be paused due to
-unforeseen delays. In this case, `start_time` should be set to `null`
-and `countdown_pause_time` to the number of seconds left to count down.
-The `countdown_pause_time` may change to indicate approximate delay.
+The typical use of `countdown_pause_time` is when a contest director wishes to pause the countdown to the start of a contest.  For example, this may occur because of technical
+issues or to make an announcement.  When the contest countdown is paused, the value of `countdown_pause_time` should be set to the expected time remaining before the start of the contest after the pause is lifted.
+In this case, the `start_time` must be set to `null`, thereby, setting an undefined `start_time` (the `start_time` is undefined since
+the length of the pause may be unknown).
+The `countdown_pause_time` may be changed to indicate the approximate delay until the contest starts.
```
niemela, line 766: This is still *slightly* inconsistent (that's a strong word though) with the description. If the value is indeed "The number of seconds left when the countdown to the contest start was paused" then it couldn't be updated. 

Maybe we should say that this is "the amount of time left to the contest start when the contest is countdown is restarted. Typically set to the current amount of time remaining when countdown is paused"?

niemela (changes requested):

Contest_API.md
johnbrvc: I don't think the last part, "_Typically set to the current amount of time remaining when the countdown is paused_" adds anything except some confusion.  I had to read it a few times before I understood it, and I know what it is supposed to be.  In fact, after a conversation with @deboer-tim yesterday, the pause time is almost always NOT the time left when the contest was paused.  It's usually some "rounded" time close to the start of the contest.  The first part is better though...  So, perhaps something like this:
`When the contest countdown is paused, the value of countdown_pause_time should be set to the amount of time left to the expected contest start when then contest countdown is restarted.  (maybe "resumed" instead of "restarted"?)
`

niemela: > the pause time is almost always NOT the time left when the contest was paused

I don't think that's correct. The `countdown_pause_time` is almost always (as in: I can't think of a single instance when it has not been) the time left at the time it was paused. It is not uncommon though that this time is changed to something rounded, possibly very shortly after.

> perhaps something like this:
> When the contest countdown is paused, the value of countdown_pause_time should be set to the amount of time left to the expected contest start when then contest countdown is restarted. (maybe "resumed" instead of "restarted"?)

I wanted to avoid the implication that you have to know this at the time of pausing (because you would typically never know that).

deboer-tim: As Fredrik said, the pause time is almost always set today by just 'stopping the clock', e.g. 3:05.57 remaining. Then after some announcements, it's often set to something like a 2:00 countdown, then a bit later the countdown will be resumed, by picking a time 2:00 in the future (rounded to the nearest second).

But all of this is based on a UI that was built for how we wanted to run at World Finals. There's no reason it couldn't pause at exactly 5:00, resume from 3:05.57, increase 'back' to 10:00, start at some arbitrary ms wall clock time, or "we're >4 minutes late already, lets just reschedule for 9:05am". It's purely a 'presentation value' to give the audience some idea of when the (currently unscheduled) contest might start. So, I think Fredrik is right that we shouldn't bake too many requirements or expectations into the spec based on how we currently use it.

Contest_API.md
```diff
@@ -749,7 +749,7 @@ Properties of contest objects:
 | name                         | string          | Short display name of the contest.
 | formal\_name                 | string ?        | Full name of the contest. Defaults to value of `name`.
 | start\_time                  | TIME ?          | The scheduled start time of the contest, may be `null` if the start time is unknown or the countdown is paused.
-| countdown\_pause\_time       | RELTIME ?       | The amount of seconds left when countdown to contest start is paused. At no time may both `start_time` and `countdown_pause_time` be non-`null`.
+| countdown\_pause\_time       | RELTIME ?       | The amount of time left when the countdown to the contest start was paused, if the contest is paused, otherwise `null`.
```
eldering, line 752:
```suggestion
| countdown\_pause\_time       | RELTIME ?       | The amount of time left when the countdown to the contest start was paused, if the contest countdown is paused, otherwise `null`.
```

Contest_API.md
```diff
@@ -759,11 +759,11 @@ Properties of contest objects:
...
 | location                     | LOCATION ?      | Location where the contest is held.
 
-The expected/typical use of `countdown_pause_time` is that once a
-`start_time` is defined and close, the countdown may be paused due to
-unforeseen delays. In this case, `start_time` should be set to `null`
-and `countdown_pause_time` to the number of seconds left to count down.
-The `countdown_pause_time` may change to indicate approximate delay.
+The typical use of `countdown_pause_time` is when a contest director wishes to pause the countdown to the start of a contest.  For example, this may occur because of technical
+issues or to make an announcement.  When the contest countdown is paused, the value of `countdown_pause_time` should be set to the expected time remaining before the start of the contest after the pause is lifted.
+In this case, the `start_time` must be set to `null`, thereby, setting an undefined `start_time` (the `start_time` is undefined since
```
eldering, line 764: Clarify that we're referring to the `countdown_pause_time` attribute, not to the example use case in the previous sentences:
```suggestion
If `countdown_pause_time` is set, then the `start_time` must be set to `null`, thereby, setting an undefined `start_time` (the `start_time` is undefined since
```

niemela (approved):

PR #176, opened 2024-12-07
Use types instead of groups for 'types of endpoints'

niemela:
...especially since "group" is an endpoint.


niemela:
(This should be backported to earlier versions as well.)

eldering (approved):

deboer-tim (approved):

PR #179, opened 2024-12-09
Add country subdivisions to organizations

### LINKED ISSUE #175: Add country subdivision to organization
Add the following properties to the organizations endpoints:

- `country_subdivision` - `string ?`- ISO 3166-2 code of the organization's country subdivision.
- `country_subdivision_flag` - `array of FILE ?` - Flag of the country subdivision. A server is recommended to provide flags of size around 56x56 and 160x160. Only allowed mime types are image/*.  

This would be used e.g. at some NA contests that show state flags (i.e. country subdivisions), e.g.: https://nasouth24.kattis.com/contests/nasouth24d1/standings .

niemela:
Fixes #175 .

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1273,7 +1273,11 @@ Returned data:
 
 ```json
 [{"id":"inst123","icpc_id":"433","name":"Shanghai Jiao Tong U.","formal_name":"Shanghai Jiao Tong University"},
- {"id":"inst105","name":"Carnegie Mellon University","country":"USA",
+ {"id":"inst105","name":"Carnegie Mellon University","country":"USA","country_subdivision":"US-PA",
+  "country_flag":[{"href":"http://example.com/api/contests/wf14/flags/USA/56px","filename":"56px.png","mime":"image/png","width":56,"height":56},
+          {"href":"http://example.com/api/contests/wf14/flags/USA/160px","filename":"160px.png","mime":"image/png","width":160,"height":160}],
+  "country_subdivision_flag":[{"href":"http://example.com/api/contests/wf14/flags/US-PA/56px","filename":"56px.png","mime":"image/png","width":56,"height":56},
+          {"href":"http://example.com/api/contests/wf14/flags/US-PA/160px","filename":"160px.png","mime":"image/png","width":160,"height":160}],
   "logo":[{"href":"http://example.com/api/contests/wf14/organizations/inst105/logo/56px","filename":"56px.png","mime":"image/png","width":56,"height":56},
           {"href":"http://example.com/api/contests/wf14/organizations/inst105/logo/160px","filename":"160px.png","mime":"image/png","width":160,"height":160}]
  }
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1247,21 +1247,21 @@ organizations (i.e. return any organization\_ids).
 
 Properties of organization objects:
 
-| Name               | Type            | Description
-| :----------------- | :-------------- | :----------
-| id                 | ID              | Identifier of the organization.
-| icpc\_id           | string ?        | External identifier from ICPC CMS.
-| name               | string          | Short display name of the organization.
-| formal\_name       | string ?        | Full organization name if too long for normal display purposes.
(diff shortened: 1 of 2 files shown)


Contest_API.md
```diff
@@ -1247,19 +1247,21 @@ organizations (i.e. return any organization\_ids).
...
+| Name                       | Type            | Description
+| :------------------------- | :-------------- | :----------
+| id                         | ID              | Identifier of the organization.
+| icpc\_id                   | string ?        | External identifier from ICPC CMS.
+| name                       | string          | Short display name of the organization.
+| formal\_name               | string ?        | Full organization name if too long for normal display purposes.
+| country                    | string ?        | [ISO 3166-1 alpha-3 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) of the organization's country.
+| country\_flag              | array of FILE ? | Flag of the country. A server is recommended to provide flags of size around 56x56 and 160x160. Only allowed mime types are image/\*.
+| country\_subdivision       | string ?        | [ISO 3166-2 code](https://en.wikipedia.org/wiki/ISO_3166-2) of the organization's country subdivision.
+| country\_subdivision\_flag | array of FILE   | Flag of the country subdivision. A server is recommended to provide flags of size around 56x56 and 160x160. Only allowed mime types are image/\*.
```
nickygerritsen, line 1259: Maybe add it to the example below?

nickygerritsen (dismissed):

Contest_API.md
johnbrvc, line 1259: Shouldn't the subdivision flag array be: `array of FILE ?`  that is, shouldn't it be optional?

Contest_API.md
```diff
@@ -1247,19 +1247,21 @@ organizations (i.e. return any organization\_ids).
...
-| logo               | array of FILE ? | Logo of the organization. A server must provide logos of size 56x56 and 160x160 but may provide other sizes as well. Only allowed mime types are image/\*.
+| Name                       | Type            | Description
+| :------------------------- | :-------------- | :----------
+| id                         | ID              | Identifier of the organization.
+| icpc\_id                   | string ?        | External identifier from ICPC CMS.
+| name                       | string          | Short display name of the organization.
+| formal\_name               | string ?        | Full organization name if too long for normal display purposes.
+| country                    | string ?        | [ISO 3166-1 alpha-3 code](https://en.wikipedia.org/wiki/ISO_3166-1_alpha-3) of the organization's country.
+| country\_flag              | array of FILE ? | Flag of the country. A server is recommended to provide flags of size around 56x56 and 160x160. Only allowed mime types are image/\*.
+| country\_subdivision       | string ?        | [ISO 3166-2 code](https://en.wikipedia.org/wiki/ISO_3166-2) of the organization's country subdivision.
```
eldering, line 1258: Maybe add for clarification:
```suggestion
| country\_subdivision       | string ?        | [ISO 3166-2 code](https://en.wikipedia.org/wiki/ISO_3166-2) of the organization's country subdivision (like state or province).
```

niemela: Added example matching the Wikipedia page.

johnbrvc (approved):

eldering (commented):
Do we want a combination like
```
"country":"NLD","country_subdivision":"US-PA"
```
to be legal? If so, I think we should explicitly mention that it's possible, if not, we should add this constraint. Or say that you SHOULD NOT do this.

niemela:
> Do we want a combination like
> 
> ```
> "country":"NLD","country_subdivision":"US-PA"
> ```

I'm a bit torn. On the one hand, that makes no sense (so, forbid it), on the other hand, why does it matter (so, allow it).

Maybe a SHOULD is the right balance, but then I don't love adding SHOULDs (which don't actually **do** anything), for things that are common sense anyway (and doesn't really hurt anything).

As I keep saying: not every dumb thing needs to be explicitly forbidden.

deboer-tim:
I'm with @niemela on this one, I think it falls into the 'dumb / clearly wrong, but not directly harmful', so I wouldn't bother explicitly forbidding it.

That said, we could change the description to something like 'subdivision within the country' which makes it a bit more explicit?

eldering (approved):

PR #185, opened 2024-12-12
Improved grammar

### LINKED ISSUE #184: Misplaced comma in Contest State description
A bit pedantic, but:

The final sentence in the paragraph below the table of properties states:
`"That, is the following sequence of inequalities must hold:"`
It should read:
`"That is, the following sequence of inequalities must hold:"`

(Move the comma from before "is" to after "is".)

I also think in the last sentence that it should be changed from:
`"A contest that has ended, has been thawed ..."`
to
`"A contest that has ended, been thawed ..."`

(Drop the 2nd "has").

niemela:
Fix #184

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1463,7 +1463,7 @@ Properties of state objects:
 These state changes must occur in the order listed in the table above,
 as far as they do occur, except that `thawed` and `finalized` may occur
 in any order. For example, the contest may never be frozen and hence not
-thawed either, or, it may be finalized before it is thawed. That, is the
+thawed either, or, it may be finalized before it is thawed. I.e., the
 following sequence of inequalities must hold:
 
 ```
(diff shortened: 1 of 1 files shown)


johnbrvc (approved):

clevengr (approved):
Looks good to me.

Contest_API.md
```diff
@@ -1463,17 +1463,17 @@ Properties of state objects:
 These state changes must occur in the order listed in the table above,
 as far as they do occur, except that `thawed` and `finalized` may occur
 in any order. For example, the contest may never be frozen and hence not
-thawed either, or, it may be finalized before it is thawed. That, is the
+thawed either, or, it may be finalized before it is thawed. I.e., the
```
eldering, line 1466: Nitpick: I think it would be better to write `That is` than `I.e.` since not everyone (especially non-native readers) clearly know the difference between `i.e.` and `e.g.`. But I can also do a separate PR globally rewriting these.

eldering (approved):

Contest_API.md
niemela: Not sure that I agree that it should be done, but I agree that if it should be done it should be a separate PR fixing it globally (and it should also be backported)

PR #193, opened 2025-01-31
Allow multiple teams and groups in clarification destinations

johnbrvc:
Removed the `to_team_id `property from the clarifications object, and replaced it with `to_team_ids `and `to_groups_ids`.  This allows a clarification response (or announcement) to be sent to multiple teams and/or groups and of course everyone.  

Added a description of how the new properties work.

Changed the clarification examples to reflect the new properties.

Changed **Filtering** example to use `to_team_ids `since it previously used `to_team_id`.

This is a "breaking change" in that clients and servers that expect the `to_team_id` property for clarifications will no longer work properly until they are updated to use the new property(ies).

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1809,7 +1809,7 @@ Properties of clarification message objects:
 | time           | TIME             | Time of the question/reply.
 | contest\_time  | RELTIME          | Contest time of the question/reply.
 
-The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`.  If `from_team_id` and both `to_team_ids` and `to_group_ids` are `null`, then the clarification is sent to all teams.  Note that if `from_team_id` is not `null`, then both `to_team_ids` and `to_group_ids` must be `null`. That is, teams cannot send messages to other teams or groups.
+The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`.  A clarification is sent to all teams if `from_team_id` is null and the union of `to_team_ids` and `to_group_ids` is empty.  Note that if `from_team_id` is not `null`, then both `to_team_ids` and `to_group_ids` must be `null`. That is, teams cannot send messages to other teams or groups.
 
 #### Modifying clarifications
 
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -594,8 +594,8 @@ Note that all results returned from endpoints:
 Endpoints that return a JSON array must allow filtering on any
 property with type ID (except the `id` property) by passing it as a
 query argument. For example, clarifications can be filtered on the
-recipient by passing `to_team_id=X`. To filter on a `null` value,
-pass an empty string, i.e. `to_team_id=`. It must be possible to
+recipient by passing `to_team_ids=X,Y`. To filter on a `null` value,
+pass an empty string, i.e. `to_team_ids=`. It must be possible to
 filter on multiple different properties simultaneously, with the
 meaning that all conditions must be met (they are logically `AND`ed).
 Note that filtering on any other property, including property with the type
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1800,7 +1800,7 @@ Properties of clarification message objects:
 | Name           | Type             | Description
 | :------------- | :--------------- | :----------
 | id             | ID               | Identifier of the clarification.
-| from\_team\_id | ID ?             | Identifier of [team](#teams) sending this clarification request, `null` iff a clarification sent by jury.
+| from\_team\_id | ID ?             | Identifier of the [team](#teams) sending this clarification request, `null` iff a clarification is sent by the jury.
 | to\_team\_ids  | array of ID ?    | Identifiers of the [team(s)](#teams) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
 | to\_group\_ids | array of ID ?    | Identifiers of the [ group(s)](#groups) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
 | reply\_to\_id  | ID ?             | Identifier of clarification this is in response to, otherwise `null`.
@@ -1809,9 +1809,7 @@ Properties of clarification message objects:
 | time           | TIME             | Time of the question/reply.
 | contest\_time  | RELTIME          | Contest time of the question/reply.
 
-Note that at least one of `from_team_id` and (`to_team_ids` or `to_group_ids`) has to be
-`null`. That is, teams cannot send messages to other teams or groups.  In order to send a reply
-to all teams, both `to_team_ids` and `to_group_ids` must be `null`.
+The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`.  If `from_team_id` and both `to_team_ids` and `to_group_ids` are `null`, then the clarification is sent to all teams.  Note that if `from_team_id` is not `null`, then both `to_team_ids` and `to_group_ids` must be `null`. That is, teams cannot send messages to other teams or groups.
 
 #### Modifying clarifications
 
@@ -1834,12 +1832,12 @@ exceptions:
   choose to include or exclude the `problem_id`.
 - The `post_clar` capability only has access to `POST`. `id`,
   `time`, and `contest_time` must not be provided. When submitting from a
-  team account, `to_team_id` must not be provided; `from_team_id` may be
+  team account, `to_team_ids` and `to_group_ids` must not be provided; `from_team_id` may be
   provided but then must match the ID of the team associated with the request.
   When submitting from a judge account, `from_team_id` must not be provided.
   In either case the server will determine an `id` and the current `time` and
   `contest_time`.
-- The `proxy_clar` capability only has access to `POST`. `id`, `to_team_id`,
+- The `proxy_clar` capability only has access to `POST`. `id`, `to_team_ids`, `to_group_ids`,
   `time`, and `contest_time` must not be provided. `from_team_id` must be
   provided. The server will determine an `id` and the current `time` and
   `contest_time`.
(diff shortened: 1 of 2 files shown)


Contest_API.md
```diff
@@ -1797,19 +1797,19 @@ The following endpoints are associated with clarification messages:
...
-| problem\_id    | ID ?    | Identifier of associated [problem](#problems), `null` iff not associated to a problem.
-| text           | string  | Question or reply text.
-| time           | TIME    | Time of the question/reply.
-| contest\_time  | RELTIME | Contest time of the question/reply.
+| Name           | Type             | Description
+| :------------- | :--------------- | :----------
+| id             | ID               | Identifier of the clarification.
+| from\_team\_id | ID ?             | Identifier of the [team](#teams) sending this clarification request, `null` iff a clarification is sent by the jury.
+| to\_team\_ids  | array of ID ?    | Identifiers of the [team(s)](#teams) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
+| to\_group\_ids | array of ID ?    | Identifiers of the [ group(s)](#groups) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
```
nickygerritsen, line 1805:
```suggestion
| to\_group\_ids | array of ID ?    | Identifiers of the [group(s)](#groups) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
```

Contest_API.md
```diff
@@ -1797,19 +1797,19 @@ The following endpoints are associated with clarification messages:
...
-| reply\_to\_id  | ID ?    | Identifier of clarification this is in response to, otherwise `null`.
-| problem\_id    | ID ?    | Identifier of associated [problem](#problems), `null` iff not associated to a problem.
-| text           | string  | Question or reply text.
-| time           | TIME    | Time of the question/reply.
-| contest\_time  | RELTIME | Contest time of the question/reply.
+| Name           | Type             | Description
+| :------------- | :--------------- | :----------
+| id             | ID               | Identifier of the clarification.
+| from\_team\_id | ID ?             | Identifier of the [team](#teams) sending this clarification request, `null` iff a clarification is sent by the jury.
+| to\_team\_ids  | array of ID ?    | Identifiers of the [team(s)](#teams) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
```
nickygerritsen, line 1804: Should we actually remove the `?` and make the empty array be the replacement? If not, we should specify that if you set the value to non-null, it should (must?) not be the empty array?

Contest_API.md
```diff
@@ -594,8 +594,8 @@ Note that all results returned from endpoints:
 Endpoints that return a JSON array must allow filtering on any
 property with type ID (except the `id` property) by passing it as a
 query argument. For example, clarifications can be filtered on the
-recipient by passing `to_team_id=X`. To filter on a `null` value,
-pass an empty string, i.e. `to_team_id=`. It must be possible to
+recipient by passing `to_team_ids=X,Y`. To filter on a `null` value,
+pass an empty string, i.e. `to_team_ids=`. It must be possible to
```
eldering, line 598: This suggests a real change: instead of just requiring filtering on properies with `ID` type, we would now also require filtering on properties with type `array of ID`.

That adds a bunch of extra complexity like: if you specify `to_team_ids=X,Y`, do you want to match entries where `to_team_ids` is exactly `[X,Y]`, contains either `X` or `Y`, has `[X,Y]` as a subset, or... I don't think we should go there, and simply change this example to something else than clarifications.

Contest_API.md
eldering: Just as a side note while thinking about this: I think we've always accepted not returning a property when it's `null`. I can't find it in the current draft anymore, did we remove this?

Contest_API.md
```diff
@@ -1879,9 +1879,9 @@ Request:
 Returned data:
 
 ```json
-[{"id":"1","from_team_id":"34","to_team_id":null,"reply_to_id":null,"problem_id":null,
+[{"id":"1","from_team_id":"34","to_team_ids":null,"to_group_ids":null,"reply_to_id":null,"problem_id":null,
   "text":"May I ask a question?","time":"2017-06-25T11:59:27.543+01","contest_time":"1:59:27.543"},
- {"id":"2","from_team_id":null,"to_team_id":"34","reply_to_id":"1","problem_id":null,
+ {"id":"2","from_team_id":null,"to_team_ids":["34"],"reply_to_id":"1","problem_id":null,
```
eldering, line 1884: Would be good to have an example where there are multiple team IDs and/or a group ID.

eldering (commented):
Should it be allowed to specify team and group IDs that overlap? That is, a team that's also already part of a group?

Contest_API.md
johnbrvc: https://ccs-specs.icpc.io/2023-06/contest_api#extensibility

Contest_API.md
johnbrvc: The "OR" of all the to_team_ids  (or to_group_ds) makes the most sense.  eg.  "Any of these teams"

johnbrvc:
> Should it be allowed to specify team and group IDs that overlap? That is, a team that's also already part of a group?

I had modified the description to say: "The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`."    Are you suggesting, perhaps it should say: "The recipients of a clarification are the union of `to_team_ids` and the teams that are members of `to_group_ids`."

Contest_API.md
johnbrvc: Agree. Will update.

eldering:
> > Should it be allowed to specify team and group IDs that overlap? That is, a team that's also already part of a group?
> 
> I had modified the description to say: "The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`." Are you suggesting, perhaps it should say: "The recipients of a clarification are the union of `to_team_ids` and the teams that are members of `to_group_ids`."

I think that's a better formulation that would be good to change. But it's not what I was trying to address.

I meant the following: let's say group `A` contains teams `X` and `Y`. Now is it legal to specify for a clarification `to_team_ids = [X,Z]` and `to_group_ids = [A]`? That is, can a team be specified in both the team and group lists, or should we explicitly forbid this? Or maybe @niemela would argue that this is overspecifying things that no-one reasonable should do? :-P

johnbrvc:
> > > Should it be allowed to specify team and group IDs that overlap? That is, a team that's also already part of a group?
> > 
> > 
> > I had modified the description to say: "The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`." Are you suggesting, perhaps it should say: "The recipients of a clarification are the union of `to_team_ids` and the teams that are members of `to_group_ids`."
> 
> I think that's a better formulation that would be good to change. But it's not what I was trying to address.
> 
> I meant the following: let's say group `A` contains teams `X` and `Y`. Now is it legal to specify for a clarification `to_team_ids = [X,Z]` and `to_group_ids = [A]`? That is, can a team be specified in both the team and group lists, or should we explicitly forbid this? Or maybe @niemela would argue that this is overspecifying things that no-one reasonable should do? :-P

I think we discussed this on the call two weeks ago.  I thought we decided that by specifying it as a "_union_" it specifically permits you have a team in the `to_team_ids `as well as be a member of a specified `to_group_ids` - because I asked the exact question you did.

Contest_API.md
niemela: > Should we actually remove the `?` and make the empty array be the replacement? If not, we should specify that if you set the value to non-null, it should (must?) not be the empty array?

I really don't like the empty list having the meaning "all".

niemela:
> I meant the following: let's say group `A` contains teams `X` and `Y`. Now is it legal to specify for a clarification `to_team_ids = [X,Z]` and `to_group_ids = [A]`? That is, can a team be specified in both the team and group lists, or should we explicitly forbid this? Or maybe @niemela would argue that this is overspecifying things that no-one reasonable should do? :-P

I would indeed argue that this is overspecifying, but I don't even think it's so unreasonable. Having overlap between the sets specified by `to_group_ids` and the set specified by `to_team_ids` is very similar to having overlaps between different sets in `to_group_ids`, and we certainly don't want to disallow that.

Contest_API.md
niemela: > I don't think we should go there, and simply change this example to something else than clarifications.

I agree, we should not add the requirement that you allow filtering on any property of type `array of ID`.

On a related note, when we say "any property with type `ID`" do we mean to include properties of type `ID ?`. We should clarify.

Contest_API.md
niemela, line 1805: As specified right now, nulls in one (but not both) of `to_team_ids` and `to_group_ids` is either illegal (because it must be `null` **iff** reply to all), (or it requires the non-null value to actually specify all teams), or it could be interpreted as the union of something and all, i.e. all. I don't think either of those are what we want.

It should be allowed to specify only `to_team_ids` (i.e. having `to_group_ids` as `null`) and that should mean that only the teams listed will get the message. Similar for only `to_group_ids`,

niemela (changes requested):

Contest_API.md
johnbrvc: Updated an example to include 3 team ids and a group id in a response.

Contest_API.md
niemela: Discussed at meeting. Will not do.

Contest_API.md
```diff
@@ -686,7 +686,7 @@ The set of properties listed must always support
 [referential integrity](#referential-integrity), i.e. if a property with a ID 
 value referring to some type of object is present the endpoint representing
 that type of object (and its ID property) must also be present. E.g. if 
-`group_ids` is listed among the properties in the `team` endpoint object, that
+`groups` is listed among the properties in the `team` endpoint object, that
```
eldering, line 689: Sorry to nitpick, but I think this would be better reverted: we actually do have a property `group_ids` in the teams endpoint, but not `groups`. And in general that format `<type>_id(s)` is how we reference other endpoint IDs in the specification.

PR #198, opened 2025-04-14
Add honours

### LINKED ISSUE #189: Update CCS System requirements to cover high and highest honours
Specifically in the [Scoring section](https://ccs-specs.icpc.io/draft/ccs_system_requirements#scoring).

niemela:
Closes #189 

It was unclear from #189 what should happen if teams match the stated rules for both "Honorable Mention" *and* one of the other. This PR is written to say that "Honorable Mention" "wins".


nickygerritsen:
> It was unclear from #189 what should happen if teams match the stated rules for both "Honorable Mention" _and_ one of the other. This PR is written to say that "Honorable Mention" "wins".

But one of the notes said:

>  Honorable Mention is empty if lowest team solved the same number of problems as the median team.

Doesn't that mean honorable mention does **not** "win"?

niemela:
> > Honorable Mention is empty if lowest team solved the same number of problems as the median team.
> 
> Doesn't that mean honorable mention does **not** "win"?

No, that doesn't actually add any new information, it just verifies that, yes, if the normal definition of honorable mention gives you an empty set, you indeed get an empty set.

The conflict that I'm referring to is (for example) what if a team solves 1 problem less than the last bronze team and the median team solves the same as the bronze team. Then the team fits both the definition of "High Honors" and "Honorable Mention". None of the comments clarify this. Same for "Highest Honors" versus "Honorable Mention".

I made choice that "Honorable Mention" wins in both these cases. This is consistent with "Honors" versus "Honorable Mention", and also just makes sense (IMO).

Contest_Control_System_Requirements.md
```diff
@@ -841,12 +841,15 @@ according to the following:
...
-8.  The *award* of a team is:
-    1.  *gold* if rank is 1 through 4.
-    2.  *silver* if rank is 5 through 8.
-    3.  *bronze* if rank is 9 through 12+B.
-    4.  *ranked* if rank is not 1 through 12+B but the team is ranked.
-    5.  *honorable* otherwise.
+8.  The *awards* of a team are:
+    1.  `gold-medal` if rank is 1 through 4.
+    2.  `silver-medal` if rank is 5 through 8.
+    3.  `bronze-medal` if rank is 9 through 12+B.
```
johnbrvc, line 847: Did he (Bill) or did he not do away with the concept of "B" ?  Can we remove it?

niemela: That was my understanding as well. (Both, that Bill has removed "B", and that @clevengr said to keep it for now).

nickygerritsen (approved):

johnbrvc (approved):

PR #200, opened 2025-04-14
Add problem limits for memory, output and code

eldering:
As discussed in last call and triggered by a feature request to DOMjudge: https://github.com/DOMjudge/domjudge/issues/2974

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1126,6 +1126,9 @@ Properties of problem objects:
 | rgb               | string ?  | Hexadecimal RGB value of problem color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet), e.g. `#AC00FF` or `#fff`.
 | color             | string ?  | Human readable color description associated to the RGB value.
 | time\_limit       | number    | Time limit in seconds per test data set (i.e. per single run). Should be a non-negative integer multiple of `0.001`. The reason for this is to not have rounding ambiguities while still using the natural unit of seconds.
+| memory\_limit     | integer   | Memory limit in MiB enforced on a submission.
+| output\_limit     | integer   | Limit in MiB on what the submission can write both to `stdout` and `stderr`. If a submission produces more output, a CCS should fail the submission or ignore output beyond this limit.
+| code\_limit       | integer   | Limit in KiB on submissions for this problem. Submissions that are larger should be rejected by a CCS.
 | test\_data\_count | integer   | Number of test data sets.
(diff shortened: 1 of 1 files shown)


nickygerritsen (dismissed):

niemela (approved):

Contest_API.md
```diff
@@ -1126,6 +1126,9 @@ Properties of problem objects:
 | rgb               | string ?  | Hexadecimal RGB value of problem color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet), e.g. `#AC00FF` or `#fff`.
 | color             | string ?  | Human readable color description associated to the RGB value.
 | time\_limit       | number    | Time limit in seconds per test data set (i.e. per single run). Should be a non-negative integer multiple of `0.001`. The reason for this is to not have rounding ambiguities while still using the natural unit of seconds.
+| memory\_limit     | integer   | Memory limit in MiB enforced on a submission.
+| output\_limit     | integer   | Limit in MiB on what the submission can write both to `stdout` and `stderr`. If a submission produces more output, a CCS should fail the submission or ignore output beyond this limit.
```
johnbrvc, line 1130: We should pick one: fail submission or ignore output, or leave them off altogether like we do  for `time_limit` and `memory_limit `(that is, we don't specify what happens for time_limit exceeded, but we "know" what happens).  I think if output limit is exceeded, it should fail, even for `stderr`.

deboer-tim (approved):
It's unfortunate that normal sizes are in MiB for one and KiB for the other (ie would have been nicer to use a consistent size integer), but I guess that's to be expected.

johnbrvc:
I think there are  few other places where a problem object is used: Notifications and Event Feed.  Maybe those should have the new properties added as well?

Contest_API.md
niemela: More importantly (?) This behavior should not be **specified** in the Contest API (but it can be described). I believe it is specified in the problem format, and maybe that's enough, but otherwise it should probably go in the CCS reqs?

johnbrvc:
> > I think there are few other places where a problem object is used: Notifications and Event Feed. Maybe those should have the new properties added as well?
> 
> Are you referring to the examples? If so, yes, that sounds like a good idea.
Yes.

Contest_API.md
```diff
@@ -1126,6 +1126,9 @@ Properties of problem objects:
...
+| code\_limit       | integer   | Limit in KiB on submissions for this problem. Submissions that are larger should be rejected by a CCS.
```
deboer-tim, line 1131: We don't use 'code' anywhere else - maybe 'submission_limit'?

Issue #202, opened 2025-05-25
Support dark/light images


nickygerritsen:
We run into issues with logos and banners a lot where they only work great on either a dark or light background.

I think it would be great if we have a way of having image 'versions' or something, where we can say they work on a light or dark background and we can support both at the same time.

eldering:
>     * Add an additional property to the file reference object, maybe `variant`? (Not a _great_ name, but I don't think we should use `type`, and most other ideas I thought of were IMO worse: "comment", "context", "appearance", "hint", "usage", "theme")

I think `variant` (and most of the other options) are too generic and would easily clash with something else in the future, and also not clearly express the intent of this property. What about `color-mode` referring to light/dark *mode*, or something similar?

niemela:
> But then it feels like this `variant` property may become a dumping ground of any tag/label/description that anyone finds useful.

Yes, it could.

I would rather have one tag/label/description field then specifying multiple (but more specific) such fields.

> Then does it become a comma-separated list or such if there are multiple non-exclusive tags/labels besides light/dark? I feel it doesn't help with clarity of how to use this field and what you can expect there.

We did discuss this at the meeting. This problem appears even for the main specific use case (dark/light images) that started this. In reality it is certainly the case that some images work only on light background, some only on dark, while some work perfectly fine on both. There is also the fact we might not yet have verified whether they work on light and/or dark yet (this can be avoided at large high resource contests like the WF, but in practice at smaller contest it could definitely be the case). How should we handle this with the `variant` property? (possibly using a different name, but let's ignore that for now).

I'm leaning towards something like this. Setting the property to "dark" means that it works well on dark backgrounds, same (but opposite) for "light". With this semantics, an image known to work on both should have both values, and an image that is *unknown* whether it works on either (or in a system that does not handle or know about this) should have neither. 

There are two good(-ish) ways to support multiple values. Either we can say that every file reference can have only one, and if you need a file to be available under many you have many file references to the same file. The other is to make `variant` an `array of string`.

> > _If_ we want to be more specific, wouldn't `theme` or maybe `style` be better?
> 
> I find that still a bit vague: for both one might expect values such as "modern", "shiny", "rounded", etc. So I'd rather make the name a bit longer and more explicit.

[...]

PR #205, opened 2025-06-09
Default filenames

### LINKED ISSUE #203: Expected filenames for CPF should be specified
Specifically a image of size A time B is expected to be called `<file>.AxB.<ext>`.

This is even more needed if we do #202 .

deboer-tim:
Updated examples to include '.' after the property name as we've been doing for years, and added a section on default filenames.

Fixes #203.

### DIFF
--- a/Contest_Package.md
+++ b/Contest_Package.md
@@ -63,6 +63,37 @@ and `contest/<filename>` respectively, and files referenced in
 Note that the API specification requires that filenames are unique within
 endpoint objects, so this is always possible.
 
+#### Default filenames
+
+In some situations you may want file references to be found on disk
+without directly specifying them in a json file, for instance to keep
+json files simple or when providing organization logos or team photos
+manually or from different systems. Files that use the default
+filename pattern must be loaded in addition to any file references
+found in the json files.
+
+The default filename pattern is
+`<endpoint>/<id>/<property>(.variant).<extension>`, where:
+
+- `<id>` is the ID of the endpoint object the reference is in.
+- `<property>` is the property of the object, e.g. `logo`.
+- `(.<variant>)` is an optional variant. Must be used when there are
+  multiple files to avoid collisions.
+- `<extension>` is a file extension corresponding to the mime type.
+
+For images, the supported file extensions are:
+
+| Mime type     | Extension(s)
+| ------------- | ---------
+| image/png     | png
+| image/jpeg    | jpg, jpeg
+| image/svg+xml | svg
+
+For images, the variant should be the width x height of the image,
+e.g. logo.56x56.png.
+
+#### Hrefs
+
 It is not required that a URL specified in an href is always valid.
 Specifically, in many cases the contest is running in a local network that is
 taken down after the contest, and in this case the URL would definitely not
@@ -138,10 +169,10 @@ problems/problemB/problemB.pdf
 ...
 groups.json
 organizations.json
-organizations/kth.se/logo56x56.png
-organizations/kth.se/logo160x160.png
-organizations/baylor.edu/logo56x56.png
-organizations/baylor.edu/logo160x160.png
+organizations/kth.se/logo.56x56.png
+organizations/kth.se/logo.160x160.png
+organizations/baylor.edu/logo.56x55.png
+organizations/baylor.edu/logo.160x157.png
 ...
 teams.json
 teams/team-001/photo.jpg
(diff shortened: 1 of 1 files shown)


nickygerritsen (dismissed):

johnbrvc:
I have a couple of comments, and if you want to take them as a request for change, that's up to you.
1) I do not like the idea of using '.' as the separator between the _property_ and the _variant_.  It complicates parsing of the basename.  I think the only '.' should denote where the _extension_ starts.   Maybe some other symbol such as '_' or '-' instead, eg. `<endpoint>/<id>/<property>(_variant).<extension>`
2) Using the term _variant_ would not be my first choice, or 10th choice for that matter.  To me, it implies something to do with data _type_, as opposed to _attributes_ for the object in question (here, the attribute would be the dimension: 56x56).

Contest_Package.md
```diff
@@ -138,10 +169,10 @@ problems/problemB/problemB.pdf
...
 groups.json
 organizations.json
-organizations/kth.se/logo56x56.png
-organizations/kth.se/logo160x160.png
-organizations/baylor.edu/logo56x56.png
-organizations/baylor.edu/logo160x160.png
+organizations/kth.se/logo.56x56.png
+organizations/kth.se/logo.160x160.png
+organizations/baylor.edu/logo.56x56.png
+organizations/baylor.edu/logo.160x160.png
```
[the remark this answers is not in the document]
deboer-tim: Good idea, repushed original commit with this change (and line before).

deboer-tim:
> 1. I do not like the idea of using '.' as the separator between the _property_ and the _variant_.  It complicates parsing of the basename.  I think the only '.' should denote where the _extension_ starts.   Maybe some other symbol such as '_' or '-' instead, eg. `<endpoint>/<id>/<property>(_variant).<extension>`

I went with `.` because that's what we've been using (at finals and NAC), and I thought the parsing was simple enough (e.g. basename / attribute it corresponds to is always what's before the first `.`, otherwise you need to look for two different chars). I don't have a strong opinion though, curious if others have input.

> 2. Using the term _variant_ would not be my first choice, or 10th choice for that matter.  To me, it implies something to do with data _type_, as opposed to _attributes_ for the object in question (here, the attribute would be the dimension: 56x56).

I thought the definition of variant was pretty accurate :). I'm fine changing, but to what? I don't think `attributes` really works. We talked about dark / light as 'tagging' an image, maybe something like 'tags'? Or we could go with any term for now and discuss as part of the light/dark proposal.

Contest_Package.md
```diff
@@ -63,6 +63,37 @@ and `contest/<filename>` respectively, and files referenced in
...
+manually or from different systems. Files that use the default
+filename pattern must be loaded in addition to any file references
+found in the json files.
+
+The default filename pattern is
+`<endpoint>/<id>/<property>(.variant).<extension>`, where:
+
+- `<id>` is the ID of the endpoint object the reference is in.
+- `<property>` is the property of the object, e.g. `logo`.
+- `(.<variant>)` is an optional variant. Must be used when there are
```
eldering, line 80: I think this will depend on the choice made in https://github.com/icpc/ccs-specs/issues/202: for example if that would become an array of strings, then probably this should become `(.<variant>)*`. Bottom-line is that a filename should be uniquely mappable to a file reference object, I think.

Contest_Package.md
```diff
@@ -63,6 +63,37 @@ and `contest/<filename>` respectively, and files referenced in
...
+For images, the supported file extensions are:
+
+| Mime type     | Extension(s)
+| ------------- | ---------
+| image/png     | png
+| image/jpeg    | jpg, jpeg
+| image/svg+xml | svg
+
+For images, the variant should be the width x height of the image,
+e.g. logo.56x56.png.
```
eldering, line 93: Aren't right now the width x height separate properties? So I'd change this to something like:
```suggestion
For images, the property (e.g. `logo`) should be followed by `.<W>x<H>` where W and H are the width and height of the image in pixels, then followed by any optional variants,
e.g. `logo.56x56.light.png`.
```

Contest_Package.md
deboer-tim: I agree - but this PR should be simple / match what we've been doing, and leave #202 to extend it.

nickygerritsen (approved):

niemela (approved):

PR #207, opened 2025-07-05
Clarify how baseurl and absolute/relative URLs interact

eldering:
This is driven by https://github.com/DOMjudge/domjudge/issues/2378 where DOMjudge had issues shadowing PC^2 because PC^2 returned absolute URLs in the file href attribute, while these should have been relative URLs (without a `/` prefix).

Also note that if a baseurl does not end with a `/`, then according to https://www.ietf.org/rfc/rfc2119.txt, the last path component must be removed when constructing a final URL with a relative URL.


eldering:
Maybe we should explicitly state that a baseurl must always end with a slash, or that it should be implicitly added always?

nickygerritsen (approved):

deboer-tim:
Many programming languages make it easy to respond to http://xyz and http://xyz/ with the same response (hey, we're just trying to be nice to the user who might have forgotten a slash). Likewise the baseURL is often printed on paper, shouted across a room, or copy/pasted - i.e. easy to miss a slash.

I don't have a concern with this change, but as per Nicky's comment I don't think it avoids the problem, it just makes it clear afterwards who was wrong. The fact that we had the examples 'wrong' is just another example of how easy it is to miss it.

I'd prefer we were more explicit on expectations, e.g. 'The contest baseurl must always end in a /. Clients should automatically correct this if they were misconfigured'.

Contest_API.md
```diff
@@ -268,14 +268,20 @@ Properties for file reference objects:
...
 ```json
   "href":"https://example.com/api/contests/wf14/submissions/187/files"
   "href":"contests/wf14/submissions/187/files"
+  "href":"/api/contests/wf14/submissions/187/files"
 ```
 
+The trailing slash in the `baseurl` is important! If `baseurl` is
+<https://example.com/api/foobar>, then the relative URL
+`contests/wf14/submissions/187/files` *also* resolves to
+`https://example.com/api/contests/wf14/submissions/187/files`.
```
nickygerritsen, line 283: What about:

```suggestion
Note that the `baseurl` always ends in a slash.
```

and move it to around line 51.

eldering: Updated.

niemela (dismissed):

deboer-tim (dismissed):
'The trailing slash in the baseurl is required when combining ...' reads like the baseurl only needs a slash when combining, and I don't think we really need the example to show why. Be a bit more explicit with just 'The baseurl must end in a slash so that relative URLs are resolved correctly.'?

Giving my +1 as I don't feel that strongly, this is better than before, and fixes an important point of failure. 👍🏼

eldering:
> 'The trailing slash in the baseurl is required when combining ...' reads like the baseurl only needs a slash when combining, and I don't think we really need the example to show why. Be a bit more explicit with just 'The baseurl must end in a slash so that relative URLs are resolved correctly.'?

I'd really prefer to keep the example there to make it clear what goes wrong: I think that's not completely obvious if you don't know the details of  the RFC.

Happy with your slightly stronger suggestion though, so adding it.

PR #212, opened 2025-07-14
Specify explicitly that nullable IDs can be filtered on

eldering:
Also explicitly mention that filtering only needs to be supported on properties listed in the access endpoint.

Adresses the "on a related note" in https://github.com/icpc/ccs-specs/pull/193#discussion_r1938694646

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -588,8 +588,9 @@ Note that all results returned from endpoints:
 ### Filtering
 
 Endpoints that return a JSON array must allow filtering on any
-property with type ID (except the `id` property) by passing it as a
-query argument. For example, clarifications can be filtered on the
+property specified in the [Access](#access) endpoint with type `ID` or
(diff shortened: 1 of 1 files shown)


deboer-tim:
TBH I don't see the need for this change, to me ID and ID? have the same type. I.e. being optional doesn't change the type. I am not opposed to it if that's confusing someone else though.

The reference to /access actually confuses me though - isn't it obvious enough that you can't filter by something you can't see or doesn't exist? I'm not sure what we're trying to clarify here, but maybe there's a simpler way to say it.

Seeing this also reminded me of the 'except the `id` property'. I know it isn't being changed here, but I'm not sure if we were trying to save someone work to implement, and it isn't a great way to do it, but it is an odd restriction to make.

niemela (dismissed):

niemela (approved):

PR #225, opened 2025-09-09
Broadcast clarification responses

### LINKED ISSUE #221: Broadcast clarification responses
We ran into an interesting situation in Baku:
 - a team sends a (private) clarification request
 - the jury responds with a (public) broadcast response
 
In this case the public feed shows a reply_to a question that the user can't see.

DOMjudge UI automatically adds the question to the response text, but the jury might edit it before responding. PC^2 doesn't. Both expose some of the original clarification to all teams.

So does broadcasting a response imply that the original message should now be public? Or should it be included in the text as appropriate? Should reply_to point to a clarification that you can't see?

deboer-tim:
Adds a comment to the Contest API about keeping referential integrity when the jury responds to a private message with a broadcast. (Corresponding CDS issue #1224 opened in ICPC tools repo)

Adds a comment to the CCS requirement indicating best practice is to use internet-style responses.

Fixes #221.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1820,6 +1820,9 @@ Properties of clarification message objects:
 
 The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`.  A clarification is sent to all teams if `from_team_id`, `to_team_ids` and `to_group_ids` are null.  Note that if `from_team_id` is not `null`, then both `to_team_ids` and `to_group_ids` must be `null`. That is, teams cannot send messages to other teams or groups.
 
+Clarifications between a team and the jury are typically private. If the jury responds with a broadcast to all teams, the
+reply_to_id should be removed for everyone who couldn't see the original message in order to preserve referrential integrity.
+
 #### Modifying clarifications
 
 To add a clarification, clients can use the `POST` method on the clarifications endpoint or the
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -640,7 +640,10 @@ following operations:
 - Determine, for any specific clarification request, what answer was returned to
   the team if the clar has already been answered.
 - Compose an answer to the clar and send it, along with the text of the original
-  clarification request, to the team.
+  clarification request, to the team. The best practice is to include the
+  original clarification text as an internet-style response, which allows the judge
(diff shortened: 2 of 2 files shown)


nickygerritsen (approved):

Contest_API.md
```diff
@@ -1820,6 +1820,9 @@ Properties of clarification message objects:
 
 The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`.  A clarification is sent to all teams if `from_team_id`, `to_team_ids` and `to_group_ids` are null.  Note that if `from_team_id` is not `null`, then both `to_team_ids` and `to_group_ids` must be `null`. That is, teams cannot send messages to other teams or groups.
 
+Clarifications between a team and the jury are typically private. If the jury responds with a broadcast to all teams, the
+reply_to_id should be removed for everyone who couldn't see the original message in order to preserve referrential integrity.
```
eldering, line 1824:
```suggestion
Clarifications between a team and the jury are typically private. If the jury sends a response to others than just the original team, then the
`reply_to_id` should be removed for everyone who couldn't see the original message in order to preserve referential integrity.
```

eldering: Related to my comment https://github.com/icpc/ccs-specs/issues/221: I'm not sure whether the "`reply_id` should be removed" should be `should` or `must` or entirely dropped.

Contest_API.md
```diff
@@ -1820,6 +1820,9 @@ Properties of clarification message objects:
 
 The recipients of a clarification are the union of `to_team_ids` and `to_group_ids`.  A clarification is sent to all teams if `from_team_id`, `to_team_ids` and `to_group_ids` are null.  Note that if `from_team_id` is not `null`, then both `to_team_ids` and `to_group_ids` must be `null`. That is, teams cannot send messages to other teams or groups.
 
+Clarifications between a team and the jury are typically private. If the jury sends a response to others than just the original team, then the
```
johnbrvc, line 1823: This sentence sounds awkward.  Maybe something like:  If the jury replies to the clarification and chooses to include additional recipients (teams?) or everyone, then the 'reply_to_id' should be removed...

eldering (approved):

Issue #230, opened 2025-09-15
Secret vs hidden teams


deboer-tim:
Without naming CCSes, :) at least 1 uses 'hidden' teams to mean admin teams, e.g. jury teams testing submissions before the contest, and who should not be visible to spectators. Another uses hidden teams to mean teams that are public, but shouldn't be included in the scoreboard (e.g. ineligible sponsor teams).

The spec leans more toward the latter, but regardless it might be better to have a second property to indicate the difference, or use a tag instead of true/false? Or maybe there's another heuristic like 'teams whose only members are admins or judges should not be publicly visible'.

niemela:
Seems like the consensus is:

1. add `account_id` to `submission`, and allow either of `team_id` and `account_id` to be null.
2. specify one `group` as the default one, the results for this group will be returned by `/scoreboard` when provided no `group_id`.
3. remove the `hidden` property from `team`.

@deboer-tim will create a PR for the first half (item 1).

PR #240, opened 2025-11-17
Optional team_id and account_id for submissions

deboer-tim:
Some CCSs allow judges or other non-team accounts to make submissions, e.g. to test the judging system or problem solutions prior to the start of a contest. This allows team_id to be null for these situations, and adds an optional account_id to indicate which account was used to make a submission.

I struggled if we should add more detail or say things like 'team_id is required iff submitted by a team', but these seemed weak and inconsistent with what we've done elsewhere. Let me know if you think we should add anything.

First part of #230.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1578,7 +1578,7 @@ Properties of submission objects:
 | id            | ID              | Identifier of the submission. Usable as a label, typically a low incrementing number to make it easier to validate submissions or compare submissions with a Shadow CCS.
 | language\_id  | ID              | Identifier of the [language](#languages) submitted for.
 | problem\_id   | ID              | Identifier of the [problem](#problems) submitted for.
-| team\_id      | ID ?            | Identifier of the [team](#teams) that made the submission.
+| team\_id      | ID ?            | Identifier of the [team](#teams) that made the submission. Submissions without a `team_id` cannot affect the scoreboard.
 | account\_id   | ID ?            | Identifier of the [account](#account) that made the submission.
 | time          | TIME            | Timestamp of when the submission was made.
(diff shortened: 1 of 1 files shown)


kunyavskiy:
>I struggled if we should add more detail or say things like 'team_id is required iff submitted by a team', but these seemed weak and inconsistent with what we've done elsewhere. 

Maybe good statement of this kind is "only submissions with non-null team can affect scoreboard".

niemela:
> Maybe good statement of this kind is "only submissions with non-null team can affect scoreboard".

On on hand, isn't that kinda obvious, because *how* would it affect the scoreboard. You could imagine that it gets added t all teams or no teams... and it's clearly not the first :). On the other hand, maybe that would be helpful?

eldering:
> > Maybe good statement of this kind is "only submissions with non-null team can affect scoreboard".
> 
> On on hand, isn't that kinda obvious, because _how_ would it affect the scoreboard. You could imagine that it gets added t all teams or no teams... and it's clearly not the first :). On the other hand, maybe that would be helpful?

Maybe going into the "if you do this, you're stupid" direction, but what if only an `account_id` is provided, but that acount actually belongs to a team?
I'd say it wouldn't hurt to simply say that only submissions with non-null team affect the scoreboard.

niemela:
> Maybe going into the "if you do this, you're stupid" direction, but what if only an `account_id` is provided, but that acount actually belongs to a team?

Very good point.

> I'd say it wouldn't hurt to simply say that only submissions with non-null team affect the scoreboard.

Agreed.

niemela (approved):

PR #241, opened 2025-11-18
Fix Secret vs hidden teams

### LINKED ISSUE #230: Secret vs hidden teams

niemela:
Continuation of #240.

Second half of #230

Closes #230.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -2252,6 +2252,8 @@ The following endpoint is associated with the scoreboard:
 | :------------------------- | :--------------- | :----------
 | `contests/<id>/scoreboard` | application/json | JSON object with scoreboard data as defined in the table below.
 
+The scoreboard includes all teams indicated by `main_scoreboard_group_id` in the `contest` endpoint.
+
 #### Scoreboard request options
 
 The following options can be passed to the scoreboard endpoint.
@@ -2261,8 +2263,6 @@ The following options can be passed to the scoreboard endpoint.
 By passing `group_id` with a valid group ID a scoreboard can be requested for the teams in a particular group:
 
 `contests/<id>/scoreboard?group_id=site1`
-
-Note that, when not passing a `group_id`, the teams indicated by `main_scoreboard_group` in the contest endpoint are used. 
 
 Each group scoreboard is ranked independently and contains only the teams that belong to the
 specified group. If a client wants to know 'local' vs 'global' rank it can query both the group and primary scoreboards.


Contest_API.md
```diff
@@ -772,6 +772,7 @@ Properties of contest objects:
 | scoreboard\_freeze\_duration | RELTIME ?       | How long the scoreboard is frozen before the end of the contest. Defaults to `0:00:00`.
 | scoreboard\_thaw\_time       | TIME ?          | The scheduled thaw time of the contest, may be `null` if the thaw time is unknown or not set.
 | scoreboard\_type             | string          | What type of scoreboard is used for the contest. Must be either `pass-fail` or `score`.
+| default_scoreboard_group     | ID ?            | Identifier of the group that represents the default scoreboard. If `null`, the default scoreboard contains all teams (even if there is no group containing all teams).
```
deboer-tim: I don't love the name (kinda long), but should also end in _id to match other fields like this. And I guess we should use backslashes in markup too?

Maybe just `scoreboard\_group\_id` to match the others? In one sense it is the default, but it's also overriding the default (all teams) behaviour too, so I'd lean more toward matching other property names.

I think the term 'default scoreboard' is potentially confusing too. I've never thought of the current scoreboard as the 'default' one - it's just 'the' scoreboard, or the 'primary' scoreboard when compared to group-specific ones. So I'd lean toward replacing:
`Identifier of the group that represents the default scoreboard. If `null`, the default scoreboard contains all teams (even if there...`
with something simpler like:
`Identifier of the group of teams to show on the scoreboard. If `null`, the scoreboard contains all teams.`

Contest_API.md
```diff
@@ -2262,6 +2262,8 @@ By passing `group_id` with a valid group ID a scoreboard can be requested for th
 
 `contests/<id>/scoreboard?group_id=site1`
 
+If no `group_id` is given, the group specified by `deafault_scoreboard_group` in the contest endpoint is used, or all teams if it is `null`.
```
deboer-tim: Typo in property name (and update if we change the name above).

I might be overly picky, but the current setup is that group_id is an optional parameter - you only add it when you want a specific group. To me this change makes it sound more like the user is expected to pass the group_id and we're specifying a default behaviour for when they don't. I'm ok if we want to include a mention of the default group id in the Scoreboard section, but I'd do that in the main part. The section on group_id shouldn't need to change b/c picking the same group as the default is maybe unnecessary, but isn't special in any way.

nickygerritsen:
> Ok, this "sudden removal need" seems very strange to me, but I'll definitely buy that it can and does happen.
> 
> That said, I don't think the process is that bad, it should be handled by the system anyway. Getting a 100 extra messages en the event feed doesn't feel too terrible (and again, this is only because you didn't get the config right to begin with, due to "sudden removal need").

I fully agree. Most submissions have more events than that for a single one and it will indeed be automated.

Contest_API.md
```diff
@@ -2262,6 +2262,8 @@ By passing `group_id` with a valid group ID a scoreboard can be requested for th
 
 `contests/<id>/scoreboard?group_id=site1`
 
+Note that, when not passing a `group_id`, the teams indicated by `main_scoreboard_group` in the contest endpoint are used.
```
deboer-tim: `_id` is missing from the property name.

My 2 cents - this still seems like we're waiting until the section on an optional parameter to describe what the primary behaviour is. I'd move this further up, e.g. in the main scoreboard description:
```
The scoreboard includes all teams indicated by `main_scoreboard_group_id` in the `contest` endpoint.
```

niemela: Agreed. Fixed.

Issue #222, opened 2025-09-04
Update CCS requirement to require (testcase) runs to be sent when availiable


kunyavskiy:
Context: https://github.com/pc2ccs/pc2v9/issues/1169

niemela:
> This implies that a judgement should be created when judging of the submission is started and then updated when it is finished.

Agreed. This means nothing for runs though (right?).

> Do we want to explicitly enforce this in the spec? That is, that all relevant runs must be present before the judgement's `judgement_type_id`, `end_time` and `end_contest_time` may be set?

How does this imply that runs must be present at that time?

I don't think it should be a requirement in the API spec that runs must be available immediately. Basically that would just mean that *if* you provide runs, you must provide them "on the fly", i.e. the only thing this forbids is making that available after, it's still very much allowed to not provide them at all.

Instead I think we should add a requirement in the CCS requiremens, because it actually *is* (i.e. should be) a requirement for a WF system to "stream" access to the runs.

In fact, we should try to add this into the current version. I'm adding this issue to the milestone. Discuss here and/or remove it if you disagree.

niemela:
> Not that they have to be published when ready, but I'd say it naturally implies that run data (if present) must be published before the final judgement verdict is added, because that verdict logically depends on these runs.

I don't see how it implies that? But also, if added *just* before the final judgement, it doesn't make much of a difference anyway.

> Sure, that sounds fine to me. I'd also make it a SHOULD in the spec. It just makes the API more useful if any data is published when available. Do we say anything about this for other entities?

I don't think we do. I would not be against a general statement along the lines of "all data should be made available on the API as soon as possible".

PR #244, opened 2025-12-15
Runs should (must at WFs) get published as soon as available

eldering:
Closes: #222


niemela:
> Reading this now feels odd for the Contest API.

Yeah... Maybe we could word it as more of a "A common use case for the `runs/` endpoint is the give real time updates on judging progress.  For this to be possible, the endpoint should be updated in near real time"? I.e. more informational, and less require-y.

nickygerritsen (approved):

PR #251, opened 2026-03-27
Factor out json format

### LINKED ISSUE #206: Factor out the JSON format to its own page
Move the JSON format specification from "Contest API" to a new "JSON Format". This would give us 4 documents:
- CCS System Requirements
- JSON Format
- Contest API
- Contest Package Format

"Contest API" and "Contest Package Format" would both refer (and link) to "JSON Format" and could focus on their specific business logic. This would be very small changes for "Contest Package Format", but a larger simplification for "Contest API".

niemela:
Closes #206 

- Factors out the definition of the JSON objects into a separate spec that both Contest API and Contest Package refer to.
- Adds more cross-links within and between documents.

I think this is an improvement. `Contest_API.md` is now about half as long and easier to navigate, with the extracted content living in the new `JSON_Format.md`. `Contest_Package.md` did not change much. More links are also good.

The semantics of the spec are unchanged.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -39,7 +39,7 @@ The key words "MUST", "MUST NOT", "REQUIRED", "SHALL", "SHALL NOT",
 document are to be interpreted as described in
 [RFC 2119](https://datatracker.ietf.org/doc/html/rfc2119).
 
-The interface is implemented as a HTTP REST interface that outputs
+The interface is implemented as an HTTP REST interface that outputs
 information in [JSON](https://en.wikipedia.org/wiki/JSON) format
 ([RFC 7159](https://datatracker.ietf.org/doc/html/rfc7159)). All access to the API
 must be provided over HTTPS to guard against eavesdropping on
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -904,710 +596,103 @@ Request data:
...
 | Endpoint                           | Type             | Description
 | :--------------------------------- | :--------------- | :----------
-| `contests/<id>/organizations`      | application/json | JSON array of all organizations with properties as specified by [access](#access).
-| `contests/<id>/organizations/<id>` | application/json | JSON object representing a single organization with properties as specified by [access](#access).
+| `contests/<id>/organizations`      | application/json | JSON array of [organization objects](json_format#organizations) with properties as specified by [access](#access).
+| `contests/<id>/organizations/<id>` | application/json | An [organization object](json_format#organizations) with properties as specified by [access](#access).
 
 Note that the first two endpoints must be provided if organizations are
 used. If they are not provided no other endpoint may refer to
 organizations (i.e. return any organization\_ids).
```
eldering: Same as above: remove?

eldering (dismissed):
In general looks fine to me. I've added some small comments/suggestions.

deboer-tim (dismissed):
As discussed on the call, I think we quickly get into a state where this is blocking/conflicting with other PRs. So, I'd suggest fixing any major issues/low hanging fruit quickly, but I'm ok merging at any point and creating a new issue to track individual areas where we should still fix.

PR #253, opened 2026-03-30
Paused intervals

deboer-tim:
An alternate solution to #182 to support paused/removed intervals where the contest clock is suspended.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -1533,6 +1533,7 @@ Properties of state objects:
 | thawed           | TIME ? | Time when the scoreboard was thawed (that is, unfrozen again), or `null` if the scoreboard has not been thawed. Required iff `scoreboard_freeze_duration` is present in the [contest](#contests) endpoint. Must not be set if frozen is `null`.
 | finalized        | TIME ? | Time when the results were finalized, or `null` if results have not been finalized. Must not be set if ended is `null`.
 | end\_of\_updates | TIME ? | Time after last update to the contest occurred, or `null` if more updates are still to come. Setting this to non-`null` must be the very last change in the contest.
+| paused_intervals | array of state pause objects ? | One or more periods where the contest clock was suspended.
 
 These state changes must occur in the order listed in the table above,
 as far as they do occur, except that `thawed` and `finalized` may occur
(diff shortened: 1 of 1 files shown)


Contest_API.md
```diff
@@ -1549,6 +1550,18 @@ A contest that has ended, been thawed (or was never frozen) and is finalized
 must not change. Thus, `end_of_updates` can be set once both `finalized` is set 
 and `thawed` is set if the contest was frozen.
 
+Properties of state pause objects:
+
+| Name         | Type     | Description
+| :----------- | :------- | :----------
+| start        | TIME     | The absolute time at which contest time was paused.
+| contest_time | REL_TIME | The contest time during the pause.
+| end          | TIME ?   | The time the pause ended and contest time resumed.
```
niemela: Same as above, "end_time" would be more consistent.

Contest_API.md
```diff
@@ -1549,6 +1550,18 @@ A contest that has ended, been thawed (or was never frozen) and is finalized
 must not change. Thus, `end_of_updates` can be set once both `finalized` is set 
 and `thawed` is set if the contest was frozen.
 
+Properties of state pause objects:
+
+| Name         | Type     | Description
+| :----------- | :------- | :----------
+| start        | TIME     | The absolute time at which contest time was paused.
+| contest_time | REL_TIME | The contest time during the pause.
```
niemela: The wording here is very subtle. It makes sense to call this "contest_time" because "start_contest_time" = "end_contest_time" (so = "contest_time"). The "during" implies that contest time is defined for entire duration.

What does this imply for how contest_time is defined before and after the contest? We don't say anything anywhere, but examples imply that contest_time is simply negative before the contest, and greater then the duration after.

niemela (changes requested):

Contest_API.md
```diff
@@ -1549,6 +1550,18 @@ A contest that has ended, been thawed (or was never frozen) and is finalized
...
+| end          | TIME ?   | The time the pause ended and contest time resumed.
+
+The contest clock may be paused for a scheduled period like a lunch break (which could be shared before the
+contest starts), or an unexpected situation like a fire alarm (where the end time is not known yet).
+If pauses happen retrospectively, the API source is responsible for ensuring the contest time of any subsequent objects is consistent.
```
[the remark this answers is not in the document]
niemela: Yes, that fixes this issue.

Contest_API.md
niemela: I agree with that interpretation. Also, it does not really have to be resolved in this PR.

niemela (changes requested):
I'm leaning towards closing this in favor of #268 .

That said, if not closed, we should add an update to the change log in readme.

niemela:
Discussed on our meeting today. Consensus was to go with the alternative solution #268. Closing this.

PR #256, opened 2026-04-07
Include event-feed.ndjson in Results upload

### LINKED ISSUE #255: Add event-feed.ndjson file to Results upload
The Results upload section says the following files should be placed in the contest root central repository:
```
api.json
api/logo.png
teams.json
scoreboard.json
awards.json
```

The `event-feed.ndjson` file should also be placed here as well.

johnbrvc:
Add event-feed.ndjson to required and optional files of the Results upload.

### DIFF
--- a/Contest_Package.md
+++ b/Contest_Package.md
@@ -242,10 +242,10 @@ Required files:
 - [api](json_format#api-information) (`api.json`)
 - [teams](json_format#teams) (`teams.json`)
 - [scoreboard](json_format#scoreboard) (`scoreboard.json`)
-- [event-feed](json_format#notification-object) (`event-feed.ndjson`)
 
 Optional files:
 - [awards](json_format#awards) (`awards.json`)
+- [event-feed](json_format#notification-object) (`event-feed.ndjson`)
 
 #### Example file listing
 
--- a/Contest_Package.md
+++ b/Contest_Package.md
@@ -242,6 +242,7 @@ Required files:
 - [api](json_format#api-information) (`api.json`)
 - [teams](json_format#teams) (`teams.json`)
 - [scoreboard](json_format#scoreboard) (`scoreboard.json`)
+- [event-feed](json_format#notification-object) (`event-feed.ndjson`)
 
 Optional files:
 - [awards](json_format#awards) (`awards.json`)
@@ -254,6 +255,7 @@ api/logo.png
 teams.json
 scoreboard.json
 awards.json
+event-feed.ndjson
 ```
 
 ## Example YAML files


deboer-tim (dismissed):
I don't know what the CMS currently supports (teams.json is 'duplicate' if you have the feed; could specify a bunch of other jsons instead), but since the feed includes everything and that's what we typically commit to git, +1.

Contest_Package.md
```diff
@@ -242,6 +242,7 @@ Required files:
 - [api](json_format#api-information) (`api.json`)
 - [teams](json_format#teams) (`teams.json`)
 - [scoreboard](json_format#scoreboard) (`scoreboard.json`)
+- [event-feed](json_format#notification-object) (`event-feed.ndjson`)
 
 Optional files:
 - [awards](json_format#awards) (`awards.json`)
```
niemela, line 247: I don't think it makes sense as a *required* file. Not all contests have it, and not all system produce it.

Optional, yes.

In fact, it would make *more* sense to say that awards is required, but I don't think we should do that either.

niemela (changes requested):

Contest_Package.md
johnbrvc: Agreed.  Change made.  This also addresses @deboer-tim 's comment above about including `teams.json` and it being a duplicate if the `event-feed.ndjson` is there.  Since `event-feed.ndjson` is now optional, it makes more sense to make `teams.json` required.

deboer-tim (approved):

niemela (changes requested):
I'm leaning towards closing this. See the discussion in #256 .

In any case, if we're not closing it, there should be an update to the readme as well.

johnbrvc:
> I'm leaning towards closing this. See the discussion in #256 .
> 
> In any case, if we're not closing it, there should be an update to the readme as well.

The reason I brought this up in the first place is because the _ICPC End of Contest_ procedure demands that `awards.json`, `event-feed.ndjson`, `scoreboard.json` all be pushed to the contest repo - presumably since they want these files saved someplace for posterity? history? the hell of it?  In any event, we used to place all this stuff in the `<contest>/results/{primary|shadow|cds}` folders - which we can still do.  However, I do think having the `event-feed.ndjson` as part of the "top level" saved contest information is not unreasonable - especially since it is optional.
  
Note: I am not saying that ICPC policy should dictate CCS specifications, but in this case I think it makes sense.  @deboer-tim  alluded to the fact that the event-feed has "everything" in it, so saving it in a known location makes sense, even it it will be saved (or linked to) in the less standard location of `<contest>/results/primary/event-feed.ndjson`.

niemela:
> The reason I brought this up in the first place is because the _ICPC End of Contest_ procedure demands that `awards.json`, `event-feed.ndjson`, `scoreboard.json` all be pushed to the contest repo - presumably since they want these files saved someplace for posterity? history? the hell of it? In any event, we used to place all this stuff in the `<contest>/results/{primary|shadow|cds}` folders - which we can still do. However, I do think having the `event-feed.ndjson` as part of the "top level" saved contest information is not unreasonable - especially since it is optional.

Right, this very reasonable.

> Note: I am not saying that ICPC policy should dictate CCS specifications, but in this case I think it makes sense. @deboer-tim alluded to the fact that the event-feed has "everything" in it, so saving it in a known location makes sense, even it it will be saved (or linked to) in the less standard location of `<contest>/results/primary/event-feed.ndjson`.

Right, I would argue that, conceptually (api) + teams + scoreboard + awards *is* a results upload (and is currently defined as such in the spec), and event-feed is a minimal *archive* upload (because it contains all the information).

With this view, what we are doing at the end of contest is a **Results upload** and an **Archive upload**.

niemela:
Discussed on our meeting today. Consensus was that the event-feed certainly needs to be uploaded, but it is not part of the **Results upload**.

We might add something about this to #278?

Closing this.

PR #257, opened 2026-04-08
Break out lists of known values

### LINKED ISSUE #254: Consider breaking out taxonomies into separate document
Taxonomies, or lists of "Known X" are fairly different than the spec in how and when it changes. I think it would make sense to break them out from the specs into separate documents with less strict versioning.

Lists that would make sense to do this to:
- Judgement types - kinda used by CCS reqs and Problem Package Format, definitely in Contest API. 
- Languages - exists in both Contest API and Problem Package Format, the former being a subset of the other.
- Contest short names - not used in this repo, but used by the problem and contest archive projects.
- Problem keywords - only used in Problem Package Format, but if we collect these lists somewhere (I think here makes more sense), it would make sense to have it as well.

niemela:
Closes #254 

Creates a new `Reference Data` page and moves the following to it:
- list of judgement types from JSON Format.
- list of languages from JSON Format and (eventually) Problem Package Format.
- list of contests from ICPC Problem Archive (and ICPC Contest Archive).
- (empty) list of problem keywords from Problem Package Format.

The semantics of the spec are unchanged.

### DIFF
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -402,59 +402,11 @@ Properties of a judgement type object:
 
 | Name                            | Type      | Description
 | :------------------------------ | :-------- | :----------
-| id                              | ID        | Identifier of the judgement type, a 2-3 letter capitalized shorthand, see table below.
+| id                              | ID        | Identifier of the judgement type, a 2-3 letter capitalized shorthand, see list of [allowed values](reference_data#judgement-types).
 | name                            | string    | Name of the judgement. (might not match table below, e.g. if localized).
 | penalty                         | boolean   | Whether this judgement causes penalty time. Required iff contest:penalty\_time is present.
 | solved                          | boolean   | Whether this judgement is considered correct.
 | simplified\_judgement\_type\_id | ID ?      | Identifier of this type's simplified judgement type.
-
-#### Known judgement types
-
-The list below contains standardized identifiers for known judgement types.
-These identifiers should be used by a server. Please create a pull request at
-<https://github.com/icpc/ccs-specs> when there are judgement types missing.
-
-The column **Big 5** lists the "big 5" equivalents, if any. A `*` in
-the column means that the judgement is one of the "big 5".
-
-The **Translation** column lists other judgements the judgement can
-safely be translated to, if a system does not support it.
-
-| ID  | Name                                     | A.k.a.                                                   | Big 5 | Translation       | Description
-| :-- | :--------------------------------------- | :------------------------------------------------------- | :---- | :---------------- | :----------
-| AC  | Accepted                                 | Correct, Yes (YES)                                       | \*    | \-                | Solves the problem
-| RE  | Rejected                                 | Incorrect, No (NO)                                       | WA?   | \-                | Does not solve the problem
-| WA  | Wrong Answer                             |                                                          | \*    | RE                | Output is not correct
-| TLE | Time Limit Exceeded                      |                                                          | \*    | RE                | Too slow
-| RTE | Run-Time Error                           |                                                          | \*    | RE                | Crashes
-| CE  | Compile Error                            |                                                          | \*    | RE                | Does not compile
-| APE | Accepted - Presentation Error            | Presentation Error, also see AC, PE, and IOF             | AC    | AC                | Solves the problem, although formatting is wrong
-| OLE | Output Limit Exceeded                    |                                                          | WA    | WA, RE            | Output is larger than allowed
-| PE  | Presentation Error                       | Output Format Error (OFE), Incorrect Output Format (IOF) | WA    | WA, RE            | Data in output is correct, but formatting is wrong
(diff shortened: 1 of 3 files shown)


Reference_Data.md
```diff
@@ -0,0 +1,424 @@
...
+| simula       | Simula 67           |                     | .sim                            |
+| smalltalk    | Smalltalk           |                     | .st                             |
+| snobol       | Snobol              |                     | .sno                            |
+| swift        | Swift               |                     | .swift                          |
+| typescript   | TypeScript          |                     | .ts                             |
+| uiua         | Uiua                |                     | .ua                             |
+| visualbasic  | Visual Basic        |                     | .vb                             |
+| zig          | Zig                 |                     | .zig                            |
+
+## Contest Short Names
```
nickygerritsen, line 154: This does not really capture the correct title does it? It is not only about short names but also about namespace and instance. I was expecting this to say something about `contest.name` (which we call the 'short display name')

deboer-tim, line 154: Unlike the other sections there's also no 'used here' links, so it feels like we're adding reference data that's not actively used?

I suspect it is or you wouldn't have included it, but IMHO would be simpler to keep this PR to extracting from the current JSON_Format (and ok, adding problem format is small) to get it in quick, then PR this section separately.

JSON_Format.md
```diff
@@ -333,60 +333,12 @@ Properties of judgement type objects:
 
 | Name                            | Type      | Description
 | :------------------------------ | :-------- | :----------
-| id                              | ID        | Identifier of the judgement type, a 2-3 letter capitalized shorthand, see table below.
+| id                              | ID        | Identifier of the judgement type, a 2-3 letter capitalized shorthand, see list of [allowed values](reference_data#judgement-types).
```
deboer-tim, line 405: Today, if a contest wanted to use a new unique judgement type I would have expected it to work like any other 'spec extension' property: they would just go ahead and define a new type `id: MAW, name: most awesome solution`, and go run a contest. If it turned out to be useful or other tools planned to use it we'd open a PR to add it to the list.

'Allowed values' makes it sound like you're not supposed to do that. Maybe go back to 'known values' or 'expected values'? We could add a bullet in the Extensibility section too.

Reference_Data.md
niemela: I (mostly) disagree, it is about contest, including what namespace they are in, and their instances.

If your point is that we should drop "Short", then I might agree...

What do you suggest?

niemela: It is used by (and the list is actually copied from) the ICPC Problem Archive project, and the ICPC Contest Archive project. Maybe I should have added links to them here? (They are linked from the introduction though).

Splitting into multiple PRs is fine with me, I guess, although I don't see any benefit with that.

JSON_Format.md
niemela: > 'Allowed values' makes it sound like you're not supposed to do that. Maybe go back to 'known values' or 'expected values'? We could add a bullet in the Extensibility section too.

Yes, but it only makes it *sound* like it :).

I would prefer of we could find a wording that pushes users towards *not* inventing their won values, while not explicitly forbidding it. I think I agree that I went too far in the "forbidding it" direction with using "allowed values" though.

Maybe we should specify some "user assigned codes", like e.g. any 3 letter code starting with X can be used for anything? That is very easy to use if you just need a quick code for some quick and dirty reason, but also ugly enough that you probably would prefer to do a PR and add a better code if you *actually* need it?

deboer-tim: In my experience there's too much of a risk that enough people use `XJAVA` before we decide it's great but clearly should be `JAVA`, nobody wants a 'breaking' change, and it becomes the defacto standard. :)

I agree with your middle paragraph - I think we just need to find a phrase that means 'just use these and play well with others (unless it really doesn't fit your case)'. Like strongly recommended values.

Reference_Data.md
nickygerritsen: `Contest Naming` maybe?

niemela: That is certainly better.

Reference_Data.md
```diff
@@ -0,0 +1,424 @@
...
+[ICPC Contest Archive](https://github.com/icpc-contest-archive/).
+
+These registries are maintained separately from the specifications that
+reference them so that new values can be added independently of specification
+release cycles. When a value needed by an implementation is not present in a
+registry, a pull request should be opened at
+<https://github.com/icpc/ccs-specs> to add it before using an ad-hoc value,
+to ensure values remain interoperable across systems.
+
+## Judgement Types
```
eldering, line 23: This file doesn't use sentence capitalization (as proposed in the style guide and used elsewhere).

niemela:
Discussed on our meeting today. Generally the idea is approved, but improvements are needed as pointed out above and in the meeting.

PR #258, opened 2026-04-09
Allow persons to have multiple roles

### LINKED ISSUE #245: Different roles in different teams for same person
Currently one person can have only one role. 

There is at least two reasonable examples of how it can be wrong:
* A person can be coach of one team and co-coach of another team
* A person can be a coach and staff member 

It's unclear how to correctly represent this in API. One of the ways could be adding several persons with same name, but diffrent roles, but this is also questionable.

niemela:
Closes #245 

Implements the fix discussed in the issue. Also adds `coach` to account types.

### DIFF
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -848,7 +848,7 @@ Properties of an account object:
 | username          | string    | The account username.
 | password          | string ?  | The account password.
 | name              | string ?  | The name of the account.
-| type              | string    | The type of account, e.g. `team`, `judge`, `admin`, `analyst`, `staff`.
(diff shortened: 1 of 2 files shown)


deboer-tim (dismissed):
+1 to this change as it's more expressive and a step in the right direction regardless, but I think we should have a short discussion about co-coach role and person/account relationship before we merge with 'Closes #245'. (i.e. wait for Monday to merge, or remove that line before merging and we can always close issue directly or do second PR)

nickygerritsen (dismissed):

eldering (dismissed):

niemela:
Discussed on our meeting today. Merging this now.

PR #260, opened 2026-04-10
Add shirt color to team

### LINKED ISSUE #234: Add a t-shirt color to team object
It's availiable on finals, we use it in live to make a coloring of team-related things (like achivements and solved problems when team is shown). 

Note: maybe also adding text color make sense, but we don't use it right now..

niemela:
Closes #234

- Add a main/body color, as well as a secondary/text color. Both RGB and human readable text name (as with problem color).

### DIFF
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -779,6 +779,10 @@ Properties of a team object:
 | desktop          | array of FILE ?        | Streaming video of the team desktop. Only allowed mime types are video/\* or application/vnd.apple.mpegurl.
 | webcam           | array of FILE ?        | Streaming video of the team webcam. Only allowed mime types are video/\* or application/vnd.apple.mpegurl.
 | audio            | array of FILE ?        | Streaming team audio.
+| shirt\_rgb       | string ?               | Hexadecimal RGB value of the team's t-shirt body color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet) with no alpha channel, e.g. `#AC00FF` or `#fff`.
+| shirt\_color     | string ?               | Human readable body color description associated to the RGB value.
+| shirt\_text\_rgb | string ?               | Hexadecimal RGB value of the team's t-shirt text color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet) with no alpha channel, e.g. `#AC00FF` or `#fff`.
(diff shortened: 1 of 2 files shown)


nickygerritsen (dismissed):

JSON_Format.md
```diff
@@ -777,6 +777,10 @@ Properties of a team object:
 | desktop          | array of FILE ?        | Streaming video of the team desktop. Only allowed mime types are video/\* or application/vnd.apple.mpegurl.
 | webcam           | array of FILE ?        | Streaming video of the team webcam. Only allowed mime types are video/\* or application/vnd.apple.mpegurl.
 | audio            | array of FILE ?        | Streaming team audio.
+| shirt\_rgb       | string ?               | Hexadecimal RGB value of the team's t-shirt body color as specified in [HTML hexadecimal colors](https://en.wikipedia.org/wiki/Web_colors#Hex_triplet) with no alpha channel, e.g. `#AC00FF` or `#fff`.
```
eldering: What if the shirt is (partially) transparent? ;-)

eldering (dismissed):

niemela:
Discussed on our meeting today. The group preferred an alternate framing where we specify a primary and secondary color instead of a shirt and text color.

deboer-tim (commented):
I like this much better than a specific t-shirt colour. 

Just want to confirm that the secondary colour is assigned though, i.e. it's not always a fixed algorithm to find something visible on the primary (which a client could replicate)?

niemela:
> Just want to confirm that the secondary colour is assigned though, i.e. it's not always a fixed algorithm to find something visible on the primary (which a client could replicate)?

I don't understand what you are asking or suggesting? I agree that that is the intent, but  we couldn't possibly add that as a requirement ("there must not be any algorithmic relationship between the primary and secondary colors"?).

nickygerritsen (approved):

deboer-tim:
> > Just want to confirm that the secondary colour is assigned though, i.e. it's not always a fixed algorithm to find something visible on the primary (which a client could replicate)?
> 
> I don't understand what you are asking or suggesting? I agree that that is the intent, but we couldn't possibly add that as a requirement ("there must not be any algorithmic relationship between the primary and secondary colors"?).

I didn't mean that as a requirement, I just meant if there's only one colour in CMS / used today and the other is always 'derived' like `(luminance > 400) ? black : white` 😉, we should only add `rgb/color` today, and could always add a secondary later. If we're already getting two colours now then I'm good with this. 👍🏼

deboer-tim (approved):

PR #261, opened 2026-04-10
Soft best practice on ordering, and add known file tags

### LINKED ISSUE #232: Ordering or preference for multiple file references
One nit I noticed while implementing the light/dark file reference tags (#227): let's say you're configuring a contest and you have a light and a dark image, but you want a specific one of them to be used by clients that don't specify a tag. e.g. maybe the light one has a border that works better against random backgrounds. Currently, there's no way to ensure that the light image comes 'first' or is picked up by clients that don't check for tags, even if you try something like providing a third untagged image. The Contest API server could serve them in any order and you don't know how clients will pick them anyway. 

Prior to this I've heard of clients that always take the first image regardless of sizes available and ordering of them, so this is somewhat an existing issue.

I don't think we can or should create rules here, but maybe we can list a best practice about how a server should order file references?

niemela:
Closes #232 

- Adding a soft best practice of listing "most generally applicable" option first.
- Adding `desktop`, and `webcam` to list of known tags.

### DIFF
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -116,8 +116,6 @@ Known values of tags include:
 
 An image should list both values if it is suitable for multiple contexts.
 
-As a best practice, servers should list file references in order from most to least generally applicable, so that clients that do not process tags can pick the first and get a reasonable result.
-
 #### Examples
 
 ```json
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -111,8 +111,12 @@ Known values of tags include:
 
 - `light`: an image suitable for use on white or light backgrounds.
 - `dark`: an image suitable for use on black or dark backgrounds.
+- `desktop`: a screenshot or capture of a team's desktop.
+- `webcam`: a capture from a team's webcam.
 
 An image should list both values if it is suitable for multiple contexts.
+
+As a best practice, servers should list file references in order from most to least generally applicable, so that clients that do not process tags can pick the first and get a reasonable result.
 
 #### Examples
 
--- a/readme.md
+++ b/readme.md
@@ -44,6 +44,9 @@ This is the draft of some future version of the CCS specification.
   `roles` array of objects, each with a `type`, optional `title`, and
   optional `team_id`, to support persons with multiple roles.
 - Added `coach` as a supported [account](json_format#accounts) type.
+- Added `desktop` and `webcam` as known [file reference](json_format#file-reference)
+  tags, and added a best practice recommendation for ordering file references
+  from most to least generally applicable.
 
 ## References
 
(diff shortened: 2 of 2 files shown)


deboer-tim:
+1 to adding desktop and webcam as known tags.

For the best practice, this is part of the reason I opened up #232 and I'm worried there are too many questions about what 'best' means and how to implement to make it actually useful, especially with tags. If you're lucky and there's a reference with no tags, or a single svg, sure, you could serve that first. But what if the sizes are 32x32, 200x200, and 5000x5000? Or if you have light and dark? Usually one of them is better on random backgrounds, but there's no way to tell which one it is.

niemela:
> For the best practice, this is part of the reason I opened up #232 and I'm worried there are too many questions about what 'best' means and how to implement to make it actually useful, especially with tags. If you're lucky and there's a reference with no tags, or a single svg, sure, you could serve that first. But what if the sizes are 32x32, 200x200, and 5000x5000? Or if you have light and dark? Usually one of them is better on random backgrounds, but there's no way to tell which one it is.

I don't understand what you want here? Are you just saying, "this is hard", then yes, it is. What the current wording does is simply, assuming a dumb client that will not make any reasonable choices, what would you suggest that it picks, put that first. I think this is reasonable, and I also think it's the best we can do, for basically the reasons you list.

deboer-tim:
> I don't understand what you want here? Are you just saying, "this is hard", then yes, it is. What the current wording does is simply, assuming a dumb client that will not make any reasonable choices, what would you suggest that it picks, put that first. I think this is reasonable, and I also think it's the best we can do, for basically the reasons you list.

Unless we're ready to create a rough algorithm to determine 'best', my point is that we shouldn't expect API servers to implement it, and hence it's also potentially dangerous for clients to have any assumption they're going to get ok results with 'first'. It's either a really simple case / we get lucky, or you just need to know. In the weird cases it's also not fair to put any onus on the server.

Even take a simple case like logo_dark.svg and logo_light.svg - client will get the 'wrong' one first 50% of the time. I think we either need to give up and admit clients need to be non-dumb, or keep it really simple (e.g. 'if there's an untagged svg put it first'?)

niemela:
> I think we either need to give up and admit clients need to be non-dumb

I am *very* fine with that.

> or keep it really simple 

Like the current suggestion? 😄 

> (e.g. 'if there's an untagged svg put it first'?)

But what's the point of that? You can already detect untagged svg (because they are untagged... and svg). That seems entirely pointless to me.

Is your suggestion do drop the "As a best practice..." sentence, merge the rest and close #232? That would sound ok to me.

niemela:
Discussed on our meeting today. Will drop the "As a best practice..." and then merge.

PR #263, opened 2026-04-10
Add score to runs

### LINKED ISSUE #215: "runs" notification and "runs" endpoint should contain score for point scoring contests
Both @meisterT and I believe for "**score**" type contests, the "**runs**" object should contain a "**score**" property (_number_) that gives the score for the run.

niemela:
Closes #215 

Adds `score` to runs.


nickygerritsen (dismissed):

eldering (dismissed):

niemela:
Discussed on out meeting today. We *might* want to clarify exactly when this score should be available in a follow up PR, but merging this now.

PR #264, opened 2026-04-10
Allow sharing output, feedback, and diff per run

### LINKED ISSUE #208: Expose problemoutput & checkoutput with `/runs`
I know the analysts prefer to get such output for analysis and now need to rerun submissions to get this output.
I also know that exposing all this data (or even storing it) might have a huge overhead so I propose to truncate this to a certain number and either only give the beginning or the end (or just remove the middle).

I don't think this is relevant for the current spec but @nickygerritsen suggested to put it on the agenda.

niemela:
Closes #208 

Adds `team_output`, `judge_message`, and `diff` to runs`.

### DIFF
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -1031,6 +1031,9 @@ Properties of a run object:
 | time                | TIME    | Absolute time when run completed.
 | run\_time           | number  | Run time in seconds. Should be a non-negative integer multiple of `0.001`. The reason for this is to not have rounding ambiguities while still using the natural unit of seconds.
 | score               | number ?| Score for this run. Only applicable when contest:scoreboard\_type is `score`. The meaning of this score is problem dependent; do not assume the final submission score is the minimum, maximum, or sum of the run scores. Note that a per-run score is not well-defined for most runs of most problems. Servers should omit `score` when it is not meaningful for the given problem.
+| team\_output        | array of FILE ? | The output produced by the team's submission for this run.
+| judge\_message      | array of FILE ? | The feedback message produced by the output validator for this run.
+| diff                | array of FILE ? | A diff between the team's output and the judge's answer for this run.
 
 #### Examples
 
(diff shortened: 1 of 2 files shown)


JSON_Format.md
```diff
@@ -982,12 +982,20 @@ Properties of run objects:
 | time                | TIME    | Absolute time when run completed.
 | contest\_time       | RELTIME | Contest relative time when run completed.
 | run\_time           | number  | Run time in seconds. Should be a non-negative integer multiple of `0.001`. The reason for this is to not have rounding ambiguities while still using the natural unit of seconds.
+| team\_output        | array of FILE ? | The output produced by the team's submission for this run.
```
nickygerritsen, line 1034: Should we say something about mimetypes of these 3? I'm not sure since we could have teams produce PNGs or something, so maybe we shouldn't say something about mime types...

niemela: This is in fact why I did not add a mime type.

JSON_Format.md
```diff
@@ -1002,12 +1002,20 @@ Properties of a run object:
...
+| judge\_message      | array of FILE ? | The feedback message produced by the output validator for this run.
+| diff                | array of FILE ? | A diff between the team's output and the judge's answer for this run.
```
eldering, line 1036: We should clarify if this meant to be a (unified) diff output or just something that in a general sense represents a diff. I guess it's the latter.

JSON_Format.md
eldering: Somewhat related: should we also say that this `array of FILE` contains exactly/at most a single file?

niemela:
Discussed on our meeting today. Several details left as homework to the drafter, but overall positive:
- support output to file rather than just stdout. Maybe using the filename in the file reference?
- support truncated outputs, but marked as such. Maybe using tags?
- we should add file size
- specify what kind of diff. Again, maybe using tags? Or mime-types?

PR #266, opened 2026-04-10
Update `max_score` when needed

### LINKED ISSUE #218: Support problem max score to be null

niemela:
Closes #218 

Alternative to #262 

Explicitly says that `max_score` must be updated when needed (i.e. when a score higher than the current `max_score` is achieved). Now, `max_score` is always required for scoring.

### DIFF
--- a/JSON_Format.md
+++ b/JSON_Format.md
@@ -646,7 +646,7 @@ Properties of a problem object:
 | output\_limit     | integer         | Limit in MiB on what the submission can write both to `stdout` and `stderr`. If a submission produces more output, a CCS should fail the submission or ignore output beyond this limit.
 | code\_limit       | integer         | Limit in KiB on submissions for this problem. Submissions that are larger should be rejected by a CCS.
 | test\_data\_count | integer         | Number of test data sets.
-| max\_score        | number          | Maximum score. Typically used to determine scoreboard cell color. Required iff contest:scoreboard\_type is `score`.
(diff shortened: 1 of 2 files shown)


JSON_Format.md
```diff
@@ -644,7 +644,7 @@ Properties of a problem object:
 | output\_limit     | integer         | Limit in MiB on what the submission can write both to `stdout` and `stderr`. If a submission produces more output, a CCS should fail the submission or ignore output beyond this limit.
 | code\_limit       | integer         | Limit in KiB on submissions for this problem. Submissions that are larger should be rejected by a CCS.
 | test\_data\_count | integer         | Number of test data sets.
-| max\_score        | number          | Maximum score. Typically used to determine scoreboard cell color. Required iff contest:scoreboard\_type is `score`.
+| max\_score        | number          | Maximum score currently considered achievable for this problem. Used by clients e.g. to shade scoreboard cells relative to the best possible result. Required iff contest:scoreboard\_type is `score`. For problems with a known maximum this value is fixed. For problems where the maximum is not known upfront, this should be set to a reasonable initial target (e.g. 100) and updated during the contest if a submission achieves a higher score.
```
eldering, line 649: Does/should this also imply that `max_score` should/must never be adjusted downward?

niemela: No, a rejudge could cause it to go down.

niemela:
Discussed on our meeting today. The consensus was that having `max_score` (or anything in `problems/`) update (outside of exceptional circumstance) is a very bad thing.

#262 seemed to be closer to a preferred solution, work will continue there.

PR #268, opened 2026-04-10
Add removed intervals to contest state

niemela:
Alternate solution to #253 and #182.

- Adds `removed_intervals` to contest state
- Add requirement that removed intervals must be non-overlapping
- Removes `contest_time` from judgement and run

### DIFF
--- a/Contest_Control_System_Requirements.md
+++ b/Contest_Control_System_Requirements.md
@@ -342,6 +342,7 @@ It must be possible to, potentially retroactively, specify time intervals that
 will be disregarded for the purpose of scoring. The time during all such
 intervals will not be counted towards a team's penalty time for solved
 problems. Beginning and end of time intervals are given in wall-clock time.
+Time intervals must not overlap.
 
 Note that removing a time interval changes the wall-clock time when the contest
 ends, as the duration of the contest in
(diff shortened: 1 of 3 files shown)


nickygerritsen (dismissed):
This is such a simple change that it might just be what we need....

I'm not opposed (but still hate that if these intervals happen in the past we need to send updates to events, although now that doesn't contain runs and judgments anymore, so I guess only submissions and clarifications? That seems reasonable.

niemela:
> but still hate that if these intervals happen in the past we need to send updates to events, 

Yes, I agree, **but**; 1) in the typical cases where you would remove intervals in the past there shouldn't be any submissions or clarifications there, 2) this should almost never happen anyway.

> although now that doesn't contain runs and judgments anymore, so I guess only submissions and clarifications

Right.

johnbrvc:
It seems we have glossed over the naming of "paused intervals" vs. "removed intervals".    I think we should discuss this on the call.

niemela:
> It seems we have glossed over the naming of "paused intervals" vs. "removed intervals". I think we should discuss this on the call.

"Removed intervals" is what the concept is currently called in the spec. As written here this would allow "pausing".

niemela:
> "Removed intervals" is what the concept is currently called in the spec. As written here this would allow "pausing".

...meaning that if we introduce "paused intervals" we still also need "removed intervals", but if we have "removed intervals", that supports both concepts.

johnbrvc:
I must have been absent on the day it was decided to remove "`start_contest_time`" and "`end_contest_time`" from "**judgements**".  At the very least, I think this should be a separate issue with a separate PR, so that it can be discussed (unless I missed the discussion that took place...)

eldering:
I like this version best of the removed/paused interval proposals.

niemela:
Discussed on our meeting today. Will merge this.

It was brought up that having a way to specify a reason could be useful. This is (weakly) related to #270, but might also be done by some other "broadcast message" mechanism. In any, case if anything it would be a follow up PR.

PR #274, opened 2026-04-11
Split CCS Requirements into WF and generic part

### LINKED ISSUE #252: Split CCS Requirements into WF and generic part
We could consider splitting `Contest_Control_System_Requirements.md` into two separate document. One describing the (sometimes very specific, and definitely very opinionated) requirements for WF, and a more generic 'this is a "CLICS compatible" CCS' part.

It is possible that second part would be very uninteresting, because of it mostly being implied by the Contest API (and Contest Package) and Problem Package Format specs. If that is the case, then it's probably not worthwhile?

Thoughts?

### LINKED ISSUE #269: Remove reference to `logos.tar.gz`
The CCS System Requirements document still refers to `logos.tar.gz` which is from (much) older versions of the spec, and is no longer used. It should be removed.

niemela:
Closes #252 
Closes #269 

This PR splits "Contest Control System Requirements" into "Contest Control System" (description of a non-WF CLICS compliant CCS) and "World Finals CCS Requirements" (requirements of a WF CCS).

I think this split ended up very well. Both parts are easier to use, and the "basic CCS" document is actually useful.

After this change, there are some structural changes, mainly on the WF document that would make a *lot* of sense, but I avoided making more changes than needed. We can do that in a follow up.

I tried to avoid making changes to what the resulting WF requirements are, but in moving things around and cleaning it up, these things changed:

0. Introduced RFC 2119 terminology. Because it's consistent with out other standards, and also because it's just good™. This is not really a *change*, but it kinda drives other changes.
1.  System should be usable with 15 problems, strengthened to a MUST. I would argue that this is what we always meant.
2. Added requirement to warn about unrecognized fields in configuration files. This part was very underspecified, and I think adding this is very reasonable.
3. Explicitly require production of the result upload files. This *was* in the old version, but the wording was very indirect: "The final results are contained in the JSON files returned by...". This is a minimal change.
4. Network security score (arguably) increased. From "the contest network" to "a network". I would argue that this actually changes nothing, and if it does it was needed 😄.
5. Predefined clarification answers "must be provided" instead of "must be supported". It's unclear to me what "must be supported" actually means in this context (basically nothing, I would argue).
6. The configuration import field list made explicit and corrected. Previously specified by reference to the Contest API output requirements, which incorrectly included fields (time_limit, test_data_count, statement, attachments) that are actually read from problem packages rather than JSON files.
7. Remove (broken) link to logos.tar.gz. Fixes #269.

I think I made some very reasonable choices in what goes in the base CCS requirements, and which are WF specific requirements. We can certainly move things around a little bit in follow up PRs if there are small (or reasonably sized) disagreements on this.


Contest_Control_System.md
```diff
@@ -0,0 +1,851 @@
...
+Primary CCS. In particular, a CCS running in shadow mode MAY NOT require that
+it be able to communicate directly with any of the actual contest participants
+(meaning the teams, judges, or contest administrators who communicate with the
+Primary CCS). Said another way, the Shadow CCS MUST be able to perform all of
+its functions while communicating only with the Primary CCS.
+
+#### CCS Configuration
+
+A CCS running in shadow mode MUST be able to import a complete contest
+configuration as described in the [Importing Contest Configuration]
```
johnbrvc, line 751: This reference does not appear to work - it appears literally in the rendered document.

Contest_Control_System.md
```diff
@@ -0,0 +1,851 @@
...
+MUST support password assignment by reading
+[accounts.json](#importing-contest-configuration). Note that if it uses another
+mechanism, it should still read that file to import (team) account information.
+
+#### Logging Out
+
+The CCS MUST support a mechanism for each user to disconnect from the contest,
+with the effect that no other user of the computer system not possessing that
+user's authentication credentials will be able to invoke CCS functions as if
+they did possess those credentials.
```
eldering, line 106: I'd argue this should be part of the WF spec. Not blocking since that can be done in a separate PR; same for similar comments below.

Contest_Control_System.md
```diff
@@ -0,0 +1,851 @@
...
+
+### Network Security
+
+All communication between CCS modules which takes place over a network MUST be
+encrypted.
+
+### Timestamps and IDs
+
+A timestamp and an integer sequence, called an ID, are assigned to submissions
```
eldering, line 115: I think `sequence` should be removed here: a single integer is assigned to a submission, not a sequence of them (like the Fibonnaci sequence).

Contest_Control_System.md
```diff
@@ -0,0 +1,851 @@
...
+The CCS MUST support "categories" to which clarification requests can be
+assigned. A request belongs to exactly one category. In addition, the CCS MUST
+construct one category per problem (e.g. "Problem A", "Problem B", etc. for
+each problem in the contest).
+
+### Language Options
+
+For each supported language compiler or interpreter it MUST be possible to
+configure the CCS to invoke it with any of the options specified in the
+compiler or interpreter's documentation.
```
eldering, line 192: Given that this implicitly refers to WF documentation, I'd say this should also move to the WF specific part.

Contest_Control_System.md
```diff
@@ -0,0 +1,851 @@
...
+When the contest is over but not finalized, all scoreboards MUST show a warning
+that the results are not final.
+
+The CCS MUST provide a way for admins to finalize the contest. It MUST NOT be
+possible to finalize a contest if one or more of the following applies:
+
+- The contest is still running (i.e., the contest is not over).
+- There are un-judged submissions.
+- There are submissions judged as [Judging Error](#exceptional-judgments).
+- There are unanswered clarification requests.
```
eldering, line 286: I think this would also be better in the WF specific part, at least the `MUST`.

Contest_Control_System.md
```diff
@@ -0,0 +1,851 @@
...
+
+#### Team Viewing of Broadcast Messages
+
+The CCS MUST provide each team with a capability for viewing any broadcast
+messages sent by the judges (see [Issuing Broadcast Messages](#issuing-broadcast-messages)).
+
+### Notifications
+
+The CCS MUST notify teams when a judgement, clarification or broadcast message
+has been received. This notification MUST NOT steal focus.
```
eldering, line 367: Just a side note: what does it mean to notify but not steal focus?

World_Finals_CCS_Requirements.md
```diff
@@ -744,245 +348,159 @@ problem should be counted as solved, as specified in the table.
...
+   [contest.json](ccs#importing-contest-configuration)) minutes for each
+   previous submission rejected with a judgement that causes penalty time, by
+   that team on that problem, or 0 if the team has not solved the problem.
+4. A team's *total penalty time* is the sum of the penalty times for all
+   problems plus any judge added penalty time.
+5. A team's *last accepted submission* is the contest time of the problem that
+   the team solved last.
+6. The *position* of a team is determined by sorting the teams first by number
+   of problems solved (descending), then within that by total penalty time
+   (ascending), then within that by last accepted submission (ascending).
```
eldering: I think these are duplicate with those points in the general CCS specification. Maybe remove them here and just refer to those?

Contest_Control_System.md
eldering: That's because it's broken across lines.

niemela: Nice catch. Note that this error is in the original though.

eldering: Still good to fix ;-) The same for the `sequence` below.

Contest_Control_System.md
niemela: I disagree that this refers to the WF documentation. It just says, "you have to be able to configure the languages ins the ways that are possible".

Contest_Control_System.md
niemela: Agreed on the MUST. Maybe on everything?

Contest_Control_System.md
niemela: If I'm coding on the next solution in my editor, the notification of a judgement must not steal the focus from my editor which would definitely mean that the next (few) characters I intended to type does not go the editor, and might also mean that e.g. the next space I type (which is very likely) will dismiss the notification before I can actually see it.

This is something that happened in systems back in the day (early 2000s).

World_Finals_CCS_Requirements.md
niemela: Yes. Nice catch.

My thinking here was that with the updates @deboer-tim and me are working on for support of generalized scoring algorithms, this would be more general in the general CCS specification docuement. But since we don't have that yet, I though keeping a copy of this in each document would actually make sense for now.

So, a bit controversial, but I would actually like to keep it this way, for now.

Contest_Control_System.md
niemela: Agreed.

Also from the original. Let's fix in follow up.

Contest_Control_System.md
eldering: Completely agree on that part of not stealing focus. I was more thinking of what counts as notification then. Anyways, I'm closing this as it's not really relevant to this PR anyways.

World_Finals_CCS_Requirements.md
eldering: Maybe good to leave a note somehow that this duplicates things? I'd just like to make sure that we don't have two copies that go out of sync at some point.

Contest_Control_System.md
eldering: Wait, I may have misread this: I was reading it as the contest's documentation (which I think is a bit pointless if there's no explicit contest referenced, like the WFs). But on second reading, I guess this means the compiler's documentation, so e.g. all possible `gcc` options. I guess that makes sense, but we've never tried *all of them* (just the compact summary is pages long) , so maybe with some it would break?

Anyways, maybe also not really relevant for this PR then.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+permalink: /wf_requirements
+---
+# World Finals CCS Requirements
+
+## Introduction
+
+This document specifies the requirements for a Contest Control System (CCS) to
+be considered for managing the operation of the
+[ICPC World Finals](https://icpc.global). It defines the specific capabilities
+and constraints that any candidate system must provide, beyond those required
```
johnbrvc: I think maybe change "any" to "a".  eg.  ... constraints that a candidate system must provide...

This would make it more consistent with a statement made in the "Rationale" section.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+be required for running other contests.
+
+Likewise, there may be functionality which a CCS provides to meet the
+requirements of some other contest; such additional functionality will not
+disqualify the CCS from qualifying as a candidate to run the ICPC World Finals
+provided the CCS meets all the requirements listed in this document.
+
+### Overview
+
+A contest control system may be submitted to the Director of Operations for the
```
johnbrvc: I think all references to "Director of Operations" should be changed to "Technical Director" in the next two paragraphs.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+- Operations
+
+## Admin Interface
+
+### Scoreboard Freeze Message
+
+As required by the [Contest Control System](ccs#freezing-the-scoreboard) base
+specification, the CCS MUST support freezing the scoreboard. When the
+scoreboard is frozen, the exact phrase displayed MUST be:
+
```
johnbrvc: Missing the "exact phrase" that was referred to.  This is what it was in the original doc:
`The scoreboard was frozen with XX minutes remaining - submissions in the last XX minutes of the contest are still shown as pending.`

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+The CCS must have a mechanism to disable any account (either an account for a
+human user or for another system), without the need for starting or stopping
+the contest. For example, this includes [team accounts]
+(#secure-authentication) and judge accounts.
+
+### Changes To Authentication Data
```
johnbrvc: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+
+The CCS must allow user authentication credential information to be changed
+dynamically by contest administration staff while the contest is running.
+
+### Starting the Contest
```
johnbrvc: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+interval is removed, the CCS must support the capability of subsequently
+restoring the contest to the state it had prior to the removal of the time
+interval.
+
+Note that time interval removal does not affect the order in which submissions
+arrived to the CCS. If submission S<sub>i</sub> arrived before submission
+S<sub>j</sub> during a removed interval, S<sub>i</sub> must still be considered
+by the CCS to have arrived strictly before S<sub>j</sub>.
+
+#### Changing contest length
```
johnbrvc: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+
+It must be possible to change the length of the contest at any time during the
+contest.
+
+#### Adding penalty time
```
johnbrvc: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+
+It must be possible to specify, for each team, an integer, potentially negative,
+amount of penalty time to be added into that team's total penalty time.
+
+#### Ignoring submissions
```
johnbrvc: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+
+It must be possible to remove, ignore or somehow mark a submission so that it in
+no way affects the scoring of the contest.
+
+### Pausing Judging
+
+The CCS must provide some way to temporarily pause judging on a per problem
```
johnbrvc, line 201: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+basis. While judging is paused teams should still be able to make submissions
+and these submission should be shown on the scoreboard as usual.
+
+### Rejudging
```
johnbrvc, line 205: This entire section appears to be a duplicate of the same section in the general CCS requirements doc.   The only difference the paragraph at the end citing an example.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+- All submissions which have been run on a specific computer (identified in some
+  reasonable way, e.g., IP address or hostname). This requirement is only
+  applicable if the CCS uses multiple machines to run submissions.
+
+Thus, for example, it must be possible to select "all rejected submissions for
+problem B", "all Time Limit Exceeded submissions using Java for problem
+C", "all submissions between time 2013-07-01 08:00:00+00 and time 2013-07-01
+09:00:00+00 of the contest using Java", or "all submissions using C++".
+
+### Manual Override of Judgments
```
johnbrvc, line 245: This section appears to be a duplicate of the same section in the general CCS requirements doc.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+The CCS must support the ability to assign, to a single submission, an updated
+judgment chosen from among any of the allowed submission judgments as defined
+in [Judge Responses](#judge-responses).
+
+The CCS must require a separate authentication every time a judgment is changed
+manually and all such changes must be logged.
+
+### Scoreboard Display
+
+The CCS must provide a mechanism for judges to view the current scoreboard. The
```
johnbrvc: While this particular paragraph is a duplicate of what appears in the base CCS requirements doc, the second paragraph is specific to WF.  I think it's probably OK to leave both in here so as not to imply the first paragraph does not apply.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+the current (updated) scoreboard as well as the frozen scoreboard.
+
+### Freezing the Scoreboard
+
+The scoreboard must automatically freeze when the configured scoreboard freeze
+time is reached. It must also be possible to manually freeze the scoreboard at
+the current time. All submissions received after the freeze time must be
+treated as pending on a frozen scoreboard.
+
+The exact phrase displayed on the frozen scoreboard must be:
```
johnbrvc, line 256: Check this with the above referenced comment about the Scoreboard Freeze Message.  Do we need both?

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+where XX is the number of minutes remaining in the contest at the time the
+scoreboard was frozen.
+
+### Finalizing the Contest
+
+As required by the [Contest Control System](ccs#finalizing-the-contest) base
+specification, the CCS MUST support finalizing the contest. In addition, the
+following World Finals-specific requirements apply.
+
+Before finalizing the contest the value B, as used in
```
johnbrvc, line 269: I thought we did away with "B"?  This year's NAC2026 did not have a value of "B" per se, rather, it was spontaneously decided during the closing ceremony to add an extra bronze medal "just because".

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+[Scoring Data Generation](#scoring-data-generation), MUST be provided. The
+default value for B MUST be 0.
+
+If, after providing the correct and final value of B, the
+[Scoreboard](contest_api#scoreboard) and [Awards](contest_api#awards) endpoints
+contain the correct results, the admin MAY finalize the contest. These
+endpoints MUST be compared with the ones exposed by the
+[Shadow CCS](ccs#shadow-mode) and SHOULD also be manually sanity checked before
+finalizing.
+
```
johnbrvc, line 279: I'm thinking perhaps there should be a "Team Interface" section that specifics a CCS MUST provide a command line utility to make submissions.  At least 50% of the teams at WF use existing command line tools to make submissions.  The command line tool could use the CLICS API to make the submissions. (which, by the way, both PC2 and DOMjudge support at the moment - in fact, we use a variation of the same tool.)

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+## Judging
+
+### Judge Responses
+
+As required by the [Contest Control System](ccs#judge-responses) base
+specification, the CCS MUST answer each submission with a response from the
+[known judgement types](json_format#known-judgement-types) list. For World
+Finals use, the CCS MUST support exactly the following responses, which include
```
johnbrvc, line 287: This differs from the WF Rules as posted on the ICPC website: https://icpc.global/worldfinals/rules in the "Conduct of the Finals" section.  This would make it consistent with the base CCS Requirements document, which only require the big 5.

World_Finals_CCS_Requirements.md
```diff
@@ -0,0 +1,637 @@
...
+```
+
+Furthermore, the CCS MUST satisfy the following:
+
+- It MUST publish [run](contest_api#runs) objects in real-time, so that
+  ICPC-Live can track and publish judging progress.
+
+#### Access Restrictions
+
+The following access restrictions MUST apply to GETs on the API endpoints:
```
johnbrvc, line 445: Should we maybe refer to the Information Access Policy, which is as-yet-to-be finalized?

johnbrvc (commented):
I painstakingly reviewed 3 documents at the same time: original, base CCS requirements, and WF CCS requirements.  The comments made in this review summarize what I found.

World_Finals_CCS_Requirements.md
niemela: Agreed.

World_Finals_CCS_Requirements.md
niemela: Agreed.

World_Finals_CCS_Requirements.md
niemela: Nice catch. Will fix!

niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: No. Removed the other one.

World_Finals_CCS_Requirements.md
niemela: Removed duplicated parts.

World_Finals_CCS_Requirements.md
niemela: Maybe we could point to it as informational, but the very fact that it is (pretty much always) yet-to-be finalized, does not make it great in a spec. I think we should try to translate the requirements we need into here.

World_Finals_CCS_Requirements.md
niemela: Judging Error is not a rejected run, so that does not contradict the rules.

We do want Security Violation though, don't we?

World_Finals_CCS_Requirements.md
niemela: I like that, but that's definitely a new requirement. I suggest making an issue (and a PR?) for that.

> At least 50% of the teams at WF use existing command line tools to make submissions

That both surprises me, and makes me happy.

World_Finals_CCS_Requirements.md
niemela: Sounds like we didn't really do away with it then?

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

World_Finals_CCS_Requirements.md
niemela: Fixed.

Contest_Control_System.md
niemela: I suggest discussing and fixing this in a follow up though.

World_Finals_CCS_Requirements.md
johnbrvc: Well, it's been removed from all EOC procedures, and, in fact, the extra Bronze at NAC was decided AFTER finalizing the contest, not before, so it really doesn't count as the old "B".  I believe @clevengr got Bill to say no more "B", or maybe it was Jeff (not Bill) before Jeff left.  @clevengr may be able to provide more details.  I'd love to get it out of the requirements doc.

World_Finals_CCS_Requirements.md
johnbrvc: I'm not sure who makes that decision.  Is there a steering committee for that?  or is the sole decision of the Technical Directory, ahem, @clevengr

World_Finals_CCS_Requirements.md
niemela: Yes, me too, but it's an actual change, and I think it's more proper to do that in a separate PR.

World_Finals_CCS_Requirements.md
johnbrvc: Indeed, for WF49 in Baku, out of the 1896 submissions, 1134 were made using the command line utility.

World_Finals_CCS_Requirements.md
niemela: This particular detail of the WF requirements is no different than the rest. Typically this is what *we* do...

eldering (approved):

johnbrvc (approved):

Contest_Control_System.md
eldering: Addressed in https://github.com/icpc/ccs-specs/pull/290

Contest_Control_System.md
eldering: Addressed in https://github.com/icpc/ccs-specs/pull/290

PR #290, opened 2026-05-14
Fix a language and markup error

eldering:
Both were detected in https://github.com/icpc/ccs-specs/pull/274.

### DIFF
--- a/Contest_Control_System.md
+++ b/Contest_Control_System.md
@@ -112,7 +112,7 @@ encrypted.
 
 ### Timestamps and IDs
 
-A timestamp and an integer, called an ID, are assigned to submissions
+A timestamp and an ID are assigned to submissions
 and clarification requests at the time they enter the system. The following
 constraints must be enforced by the CCS:
 
--- a/Contest_Control_System.md
+++ b/Contest_Control_System.md
@@ -112,7 +112,7 @@ encrypted.
 
 ### Timestamps and IDs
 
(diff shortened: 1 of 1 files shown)


Contest_Control_System.md
```diff
@@ -112,7 +112,7 @@ encrypted.
 
 ### Timestamps and IDs
 
-A timestamp and an integer sequence, called an ID, are assigned to submissions
+A timestamp and an integer, called an ID, are assigned to submissions
```
deboer-tim: The fix is good, but technically IDs are not required to be integers: there's the regular ID regex and `Usable as a label, typically a low incrementing number`. IMHO something like `s57` would be 'non-typical' but perfectly acceptable.

Maybe something like 'A timestamp and an ID (typically a low integer), are assigned'

niemela: Or just "A timestamp and an ID, are assigned" to not repeat the suggestions slightly differently in many places?

niemela (changes requested):

Contest_Control_System.md
niemela:
```suggestion
A timestamp and an ID are assigned to submissions
```

niemela (changes requested):

niemela (approved):

deboer-tim (approved):

nickygerritsen (approved):

PR #295, opened 2026-06-16
Add Clarification Categories

johnbrvc:
1. Updates to JSON format and API to include new object for clarification categories.
2. Update clarification object to include a category_id.
3. Update notifications to include "categories".
4. Updated examples for clarifications to included more consistent dates and to include an example using category_id.

I believe I tracked down the places that needed to be updated as a result of adding this new type.  If I missed anything, LMK.  Schema will probably have to be updated at some point, but I think @eldering usually does that?

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -180,7 +180,7 @@ suitable for the client making the request, as per the following example:
 
 ```json
 {"code":403,
- "message":"Teams cannot send clarifications to another team"}
+ "message":"Teams cannot send s to another team"}
  ```
 
 ### Authentication
@@ -283,7 +283,7 @@ The endpoints can be categorized into 4 types as follows:
 
 - Metadata: [api](#api-information), [access](#access)
 - Configuration: accounts, contests, judgement-types, languages, problems,
-  groups, organizations, persons, teams, categories;
+  groups, organizations, persons, teams, clarification-categories;
 - Live data: state, submissions, judgements, runs, clarifications, awards,
   commentary;
 - Aggregate data: [scoreboard](#scoreboard), [event-feed](#event-feed).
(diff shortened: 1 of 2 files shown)


nickygerritsen (dismissed):
I do wonder if we should call it `clarification-category` instead of `category` if we ever get problem categories or something, but I'm not sure.

Contest_API.md
```diff
@@ -283,7 +283,7 @@ The endpoints can be categorized into 4 types as follows:
 
 - Metadata: [api](#api-information), [access](#access)
 - Configuration: accounts, contests, judgement-types, languages, problems,
-  groups, organizations, persons, teams;
+  groups, organizations, persons, teams, categories;
```
niemela: "categories" is to generic. Maybe "clarification-categories"?

niemela (changes requested):

Contest_API.md
johnbrvc: Yeah.  I thought about that too.  It's just that "`clarification-categories`" is soooooo long.  We can discuss on the call today.  One idea I had was to add a "`type`" property to the "**Categories**" object, and then we can create different types of categories: **clarifications** being the first.  This way, if we ever need categories for something else, the framework is there.

jbuck-cs: Decided to change to "`clarification-categories`".

johnbrvc: Branch updated with new EP name.

nickygerritsen (dismissed):

niemela (dismissed):

deboer-tim (dismissed):

nickygerritsen (dismissed):

JSON_Format.md
```diff
@@ -1094,6 +1118,7 @@ Properties of a clarification message object:
 | to\_group\_ids | array of ID ?    | Identifiers of the [group(s)](#group) receiving this reply, `null` iff a reply to all teams or a request sent by a team.
 | reply\_to\_id  | ID ?             | Identifier of clarification this is in response to, otherwise `null`.
 | problem\_id    | ID ?             | Identifier of associated [problem](#problem), `null` iff not associated to a problem.
+| category\_id   | ID ?             | Identifier of the associated [clarification category][#clarification-category], `null` iff associated to a problem
```
eldering, line 1121: Nitpick on the final dot for consistency:
```suggestion
| category\_id   | ID ?             | Identifier of the associated [clarification category][#clarification-category], `null` iff associated to a problem.
```

JSON_Format.md
```diff
@@ -1103,25 +1128,27 @@ The recipients of a clarification are the union of `to_team_ids` and `to_group_i
 Clarifications between a team and the judges are typically private. If the judges replies to a clarification and chooses to include additional recipients,
 then in order to preserve referential integrity the `reply_to_id` should be removed for everyone who couldn't see the original message.
 
+At least one of `problem_id` or `category_id` must be `null` or missing.  If both are `null` or missing, then the clarification is assumed to be a general clarification.
```
eldering, line 1131: This is inconsistent with the description of `category_id` above that says that it should be `null` IFF it is associated to a problem.

Also a side-note: given that attributes with `null` values can be left out, it's not so easy to figure out as client whether this attribute is `null` because the clarification is associated to a problem or is a general clarification, or whether clarification categories are simply not used at all (other than checking the `access` endpoint).

eldering:
>  Schema will probably have to be updated at some point, but I think @eldering usually does that?

Yes, I can do that once this is merged. Pleas remind me if I forget.

JSON_Format.md
johnbrvc: I'm not sure I understand this comment.   If `problem_id `is non-null, then you don't care about `category_id` since it must be null/missing.  If `problem_id `is null, then `category_id `may be null/missing (in which case it's a general clarification), otherwise it's the ID of the category.  What's not easy about that?

johnbrvc: I do agree, that if `category_id `is null, it does not necessarily mean that `problem_id `is non-null (that is, the clarification is associated with a problem), since it may be a general clarification.  I'll need to think about wording for that.

niemela (dismissed):

niemela (approved):

deboer-tim (approved):

PR #306, opened 2026-08-05
Markdown style updates for new mkdocs HTML generator

eldering:
This introduces no contents changes. It just makes a few changes in markdown formatting since the old Jekyll HTML generator used the commonmark dialect, while mkdocs uses something slightly different.

Once this is approved and merged, I will apply similar changes to the release branches.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -181,7 +181,7 @@ suitable for the client making the request, as per the following example:
 ```json
 {"code":403,
  "message":"Teams cannot send clarifications to another team"}
- ```
+```
 
 ### Authentication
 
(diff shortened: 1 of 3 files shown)


eldering:
Never mind: I now checked the rendered HTML and it looks fine. This was AI hallucination, so closing this. I'll check the other releases if there are any rendering issues.

eldering:
Never mind: I now checked the rendered HTML and it looks fine. This was AI hallucination, so closing this. I'll check the other releases if there are any rendering issues.

**Edit:** I was not checking carefully. https://github.com/icpc/ccs-specs/blob/gh-pages/configs/hooks.py automatically fixes this. I want to remove that, so this needs to be applied.

deboer-tim (approved):

nickygerritsen (approved):

Issue #305, opened 2026-07-27
FileReference tag inconsistencies


deboer-tim:
I noticed a couple minor things while doing some work comparing spec versions:
- Should FIleReference.tag be optional in 2026-01 and draft? Sometimes we don't know anything about an image, and 2/3 examples don't include it.
- FIleReference.tag is singular but (e.g.) new Person.roles in draft is plural. I guess tags should have been plural, but not worth 'fixing'?

deboer-tim:
I was looking at this again and confirmed:
- we've used plural whenever an array has multiple independent or different things, e.g. contest objects like teams or problems, properties like submission files, team group ids, or commentary tags.
- we used singular whenever an array had different formats or resolutions of an identical thing, e.g. organization logo or team desktop.

Given this, submission `reaction` (which can now be separate webcam and desktop videos) and file reference `tag` (which could be "light" and "high-contrast" and "new-2026") both feel wrong to me.

Reaction has been around for several spec versions and too late to change. Tag was just added, and the only implementation I'm aware of (CDS) that has been successfully tested at contests... accidentally used 'tags' as the expected property name. This is the one case where I think it's actually better to fix the spec as a minor 'oops' vs fixing the implementation.

PR #308, opened 2026-08-11
Make FileReference tags consistent in 2026-01

deboer-tim:
We added a FileReference `tag` property in 2026-01, but this isn't consistent with how we've used plural elsewhere in the specification. Inconsistent enough that the only known implementations of it so far (CDS/ICPC tools/team view) had `tags` as a typo, and was successfully tested at NAC. See #305 for more details.

Normally we should fix the implementation and not touch a closed spec, but this seems like a minor typo that's still easy to correct and less change than the sole implementation.

No direct comments on the issue so I figured I'd open a PR for feedback. :)

First half of #305, second PR would change master.

### DIFF
--- a/Contest_API.md
+++ b/Contest_API.md
@@ -273,7 +273,7 @@ Properties for file reference objects:
 | mime     | string          | Mime type of resource.
 | width    | integer ?       | Width of the image. Required for files with mime type image/\*.
 | height   | integer ?       | Height of the image. Required for files with mime type image/\*.
-| tag      | array of string | Intended usage hints (e.g. `light`, `dark` for images). No meaning must be implied or inferred from the order of the elements.
+| tags     | array of string | Intended usage hints (e.g. `light`, `dark` for images). No meaning must be implied or inferred from the order of the elements.
 
 The `href` property may be an [absolute or relative
 URL](https://datatracker.ietf.org/doc/html/rfc3986); relative URLs must be
--- a/Contest_Package.md
+++ b/Contest_Package.md
@@ -91,7 +91,7 @@ For images, the supported file extensions are:
 For images, a tag of `.<W>x<H>`, is interpreted as the file reference object
 having the `width` property set to  `<W>` and the `height` property set to
 `<H>`. Every other tag is interpreted as the file reference object having
-the value in its `tag` property array.
+the value in its `tags` property array.
 
 #### Hrefs
 
--- a/readme.md
+++ b/readme.md
@@ -28,7 +28,7 @@ Other versions of the CCS specifications are available
 These are the main changes made since the `2023-06` version:
 
    1. Add support for **post\_comment** capability (posting commentary)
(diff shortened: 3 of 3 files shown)


eldering (dismissed):
Sounds ok to me. I think we should mention this somewhere clearly, in case someone did implement the original version of the `2026-01` spec, or downloaded a copy of it somehow.

nickygerritsen (dismissed):

deboer-tim:
> Sounds ok to me. I think we should mention this somewhere clearly, in case someone did implement the original version of the `2026-01` spec, or downloaded a copy of it somehow.

Where do you think the best place to put that is? We could add it to the readme, e.g.:
 
`   2. Add property **tags** (array of string) to file reference objects (FILE) along with a description of how it may be used. (eg. hints such as light or dark for images) **This property was initially mis-published as 'tag'**`

eldering:
> > Sounds ok to me. I think we should mention this somewhere clearly, in case someone did implement the original version of the `2026-01` spec, or downloaded a copy of it somehow.
> 
> Where do you think the best place to put that is? We could add it to the readme, e.g.:
> 
> ` 2. Add property **tags** (array of string) to file reference objects (FILE) along with a description of how it may be used. (eg. hints such as light or dark for images) **This property was initially mis-published as 'tag'**`

Yes, that sounds good. Maybe we should also add it briefly to https://github.com/icpc/ccs-specs/blob/gh-pages/README.md, so that when people follow the link from https://ccs-specs.icpc.io/ they're aware that the `2026-01` spec has gotten a minor change, contrary to mentioned there that releases are stable.

deboer-tim:
> > > Sounds ok to me. I think we should mention this somewhere clearly, in case someone did implement the original version of the `2026-01` spec, or downloaded a copy of it somehow.
> > 
> > Where do you think the best place to put that is? We could add it to the readme, e.g.:
> > ` 2. Add property **tags** (array of string) to file reference objects (FILE) along with a description of how it may be used. (eg. hints such as light or dark for images) **This property was initially mis-published as 'tag'**`
> 
> Yes, that sounds good. Maybe we should also add it briefly to [`gh-pages`/README.md](https://github.com/icpc/ccs-specs/blob/gh-pages/README.md?rgh-link-date=2026-08-13T21%3A35%3A58Z), so that when people follow the link from https://ccs-specs.icpc.io/ they're aware that the `2026-01` spec has gotten a minor change, contrary to mentioned there that releases are stable.

Proposed comment added here which reset approvals, please take another look. The remaining issues are all on different branches so if we get 2 approvals here let's merge and track the rest via other PRs:

- PR #309 opened to make the equivalent change to master branch.
- PR #310 open to add a comment to `gh-pages`/README.md list of releases.

eldering (dismissed):

niemela (dismissed):

eldering (approved):

niemela (approved):

Issue #302, opened 2026-07-10
Formalize 2026-draft naming


deboer-tim:
On the April 27 and May 11 calls we discussed the fact that an API returning `draft` as the version could mean literally any version/date that's not 2023-06 or 2026-01 - i.e. you can't tell if 'draft' is trying to implement the latest proposed changes or 4 years old.

Our decision was to call the current draft `2026-draft` and use that pattern going forward since we're unlikely to release more than once a year - i.e. `<yyyy>-draft` means 'changes after the <yyyy>-<xx> version towards the following spec'. You still won't know how much of the following spec it implements or whether it's correct, but at least you know what it's working toward.

The JSON Format says: `Version of the API. For this version must be the string 'draft'. Will be of the form <yyyy>-<mm>, <yyyy>-<mm>-draft, or simply draft.` We should use a PR to confirm naming of this draft and allow `<yyyy>-draft`.

eldering:
Agreed with what was discussed, but rethinking this now, I think it would be good to pick a draft version that is clearly later than the current release `2026-01`. So I'd propose `2027-draft` instead, or if we think that we might release another version in 2026, then maybe `2026-xx-draft`?

deboer-tim:
It's probably hopelessly optimistic, but if we name it `2027-draft` and then we actually turned around a spec in under a year (e.g. `2026-11`) or even 18 months (e.g. `2027-07`) I think it's still confusing which period that draft was in.

I was starting to wonder if we should follow the original text and use `2026-02-draft` ("the draft started the month after 2026-01") or as some projects do `2026-01-next` ("the version after 2026-01").

deboer-tim:
From the meeting:
- Continue using `<yyyy>-<mm>` for releases. These are the only safe, long-term versions.
- Use `<yyyy>-<mm>-next` for 'draft work following the yyyy-mm release'.
- Use `<yyyy>-<mm>-draft` for 'final review, we intend to declare a new release yyyy-mm'.
- Remove 'draft' (these are now unknown releases).
- Tim to create PR after current merge queue cleared.

deboer-tim:
Opened #311 to update the current draft.

Should I go back to past releases (2023-06, 2026-01, etc) and update them as well? i.e. change
```
Will be of the form `<yyyy>-<mm>`, `<yyyy>-<mm>-draft`, or simply `draft`.
```
to
```
[...]

PR #311, opened 2026-09-14
Update release version naming

deboer-tim:
Updates to follow the agreed release naming:
- `<yyyy>-<mm> for` releases. These are the only safe, long-term versions.
- `<yyyy>-<mm>-next` for 'draft work following the yyyy-mm release'.
- `<yyyy>-<mm>-draft` for 'final review, we intend to declare a new release yyyy-mm'.
- Removes 'draft' (these are now unknown releases).

Part of #302.


JSON_Format.md
```diff
@@ -267,7 +267,7 @@ Properties of the provider object:
 
 ```json
 {
-   "version": "draft",
+   "version": "2026-01-next",
    "version_url": "https://ccs-specs.icpc.io/draft/contest_api",
```
johnbrvc, line 271: Is the version URL going to still just use the word "draft" in it?

deboer-tim: @eldering we probably should change the urls to match... but is that possible and how much work would that be? Changing branch name and republishing?

